PR #11, opened 2024-01-19
inflate more SIMD

folkertdev:
needs refinement, and just hardcodes x86 + AVX2.

The general gist here is that we get fewer instructions, but the same number of cycles as a safer version without this much SIMD. But that can be very instruction-set dependent. So we should have more instruction sets before we can really make this decision.


folkertdev:
I tried to improve the `offset_from_end == 1` case, but it is only faster in the micro benchmark, and actually slower in the inflate of `silesia-small.tar.gz`

PR #24, opened 2024-02-07
Implement generic CRC-32

ahomescu:
Add a naive implementation of CRC-32 based
on the C code from the PNG specification.

### DIFF
--- a/zlib-rs/src/crc32.rs
+++ b/zlib-rs/src/crc32.rs
@@ -1,0 +1,169 @@
+// Several implementations of CRC-32:
+// * A naive byte-granularity approach
+// * A word-sized approach that processes a usize word at a time
+// * A "braid" implementation that processes a block of N words
+//   at a time, based on the algorithm in section 4.11 from
+//   https://github.com/zlib-ng/zlib-ng/blob/develop/doc/crc-doc.1.0.pdf.
+
+// The binary encoding of the CRC-32 polynomial.
+// We are assuming little-endianness so we process the input
+// LSB-first. We need to use the "reversed" value from e.g
+// https://en.wikipedia.org/wiki/Cyclic_redundancy_check#Polynomial_representations.
+const CRC32_LSB_POLY: usize = 0xedb8_8320usize;
+
+const W: usize = core::mem::size_of::<usize>();
+
+// The logic assumes that W >= sizeof(u32).
+// In Rust, this is generally true.
+const _: () = assert!(W >= core::mem::size_of::<u32>());
+
+// Pre-computed tables for the CRC32 algorithm.
+// CRC32_BYTE_TABLE corresponds to MulByXPowD from the paper.
+static CRC32_BYTE_TABLE: [[u32; 256]; 1] = build_crc32_table::<256, 1, 1>();
+// CRC32_WORD_TABLE is MulWordByXpowD.
+static CRC32_WORD_TABLE: [[u32; 256]; W] = build_crc32_table::<256, W, 1>();
+
+// Work-around for not being able to define generic consts or statics
+// Crc32BraidTable::<N>::TABLE is the generic table for any braid size N.
+struct Crc32BraidTable<const N: usize>;
+
+impl<const N: usize> Crc32BraidTable<N> {
+    const TABLE: [[u32; 256]; W] = build_crc32_table::<256, W, N>();
+}
+
+// Build the CRC32 tables using a more efficient and simpler approach
+// than the combination of Multiply and XpowN (which implement polynomial
+// multiplication and exponentiation, respectively) from the paper,
+// but with identical results. This function is const, so it should be
+// fully evaluated at compile time.
+const fn build_crc32_table<const A: usize, const W: usize, const N: usize>() -> [[u32; A]; W] {
+    let mut arr = [[0u32; A]; W];
+    let mut i = 0;
+    while i < W {
+        let mut j = 0;
+        while j < A {
+            let mut c = j;
+            let mut k = 0;
+            while k < 8 * (W * N - i) {
+                if c & 1 != 0 {
+                    c = CRC32_LSB_POLY ^ (c >> 1);
+                } else {
+                    c >>= 1;
+                }
+                k += 1;
+            }
+            arr[i][j] = c as u32;
+            j += 1;
+        }
+        i += 1;
+    }
+    arr
+}
+
+fn crc32_naive_inner(data: &[u8], start: u32) -> u32 {
+    data.iter().fold(start, |crc, val| {
+        let crc32_lsb = crc.to_le_bytes()[0];
+        CRC32_BYTE_TABLE[0][usize::from(crc32_lsb ^ *val)] ^ (crc >> 8)
+    })
+}
+
+fn crc32_words_inner(words: &[usize], start: u32, per_word_crcs: &[u32]) -> u32 {
+    words.iter().enumerate().fold(start, |crc, (i, word)| {
+        let value = *word ^ (crc ^ per_word_crcs.get(i).unwrap_or(&0)) as usize;
+        value
+            .to_le_bytes()
+            .into_iter()
+            .zip(CRC32_WORD_TABLE)
+            .fold(0u32, |crc, (b, tab)| crc ^ tab[usize::from(b)])
+    })
+}
+
+pub fn crc32_naive(data: &[u8], start: u32) -> u32 {
+    let crc = !start;
+    let crc = crc32_naive_inner(data, crc);
+    !crc
+}
+
+pub fn crc32_words(data: &[u8], start: u32) -> u32 {
+    // Get a word-aligned sub-slice of the input data
+    let (prefix, words, suffix) = unsafe { data.align_to::<usize>() };
+    let crc = !start;
+    let crc = crc32_naive_inner(prefix, crc);
+    let crc = crc32_words_inner(words, crc, &[]);
+    let crc = crc32_naive_inner(suffix, crc);
+    !crc
+}
+
+pub fn crc32_braid<const N: usize>(data: &[u8], start: u32) -> u32 {
+    // Get a word-aligned sub-slice of the input data
+    let (prefix, words, suffix) = unsafe { data.align_to::<usize>() };
+    let crc = !start;
+    let crc = crc32_naive_inner(prefix, crc);
+
+    let mut crcs = [0u32; N];
+    crcs[0] = crc;
+
+    // TODO: this would normally use words.chunks_exact(N), but
+    // we need to pass the last full block to crc32_words_inner
+    // because we accumulate partial crcs in the array and we
+    // need to roll those into the final value. The last call to
+    // crc32_words_inner does that for us with its per_word_crcs
+    // argument.
+    let blocks = words.len() / N;
+    let blocks = blocks.saturating_sub(1);
+    for i in 0..blocks {
+        // Load the next N words.
+        let mut buffer: [usize; N] =
+            core::array::from_fn(|j| usize::to_le(words[i * N + j] ^ (crcs[j] as usize)));
+
+        crcs.fill(0);
+        for j in 0..W {
+            for k in 0..N {
+                crcs[k] ^= Crc32BraidTable::<N>::TABLE[j][buffer[k] & 0xff];
+                buffer[k] >>= 8;
+            }
+        }
+    }
+
+    let crc = core::mem::take(&mut crcs[0]);
+    let crc = crc32_words_inner(&words[blocks * N..], crc, &crcs);
+    let crc = crc32_naive_inner(suffix, crc);
+    !crc
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    quickcheck::quickcheck! {
+        fn naive_is_crc32fast(v: Vec<u8>, start: u32) -> bool {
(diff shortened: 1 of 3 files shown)


zlib-rs/src/crc32.rs
```diff
@@ -0,0 +1,133 @@
...
+fn crc32_words_inner(words: &[usize], start: u32, per_word_crcs: &[u32]) -> u32 {
+    words.iter().enumerate().fold(start, |crc, (i, word)| {
+        let mut word_crc = 0u32;
+        let value = *word ^ (crc ^ per_word_crcs.get(i).unwrap_or(&0)) as usize;
+        for (i, b) in value.to_le_bytes().into_iter().enumerate() {
+            word_crc ^= CRC32_WORD_TABLE[i][usize::from(b)];
+        }
+
+        // Per Listing 4 in Kadatch and Jenkins, this step
+        // is needed in case size_of(usize) < size_of(u32)
```
folkertdev, line 56: in rust, this will never happen (only 32-bit and 64-bit targets are supported). To express this assumption maybe an assert in a constant would work

```rust
// the logic assumes sizof(usize) >= sizeof(u32). In rust, this is always true
const _: () = assert!(core::mem::size_of_val(&word_crc) <= W);
```

zlib-rs/src/crc32.rs
```diff
@@ -0,0 +1,133 @@
+const CRC32_LSB_POLY: usize = 0xedb8_8320usize;
+const START_CHECKSUM: u32 = 0xffff_ffffu32;
+const FINAL_XOR: u32 = 0xffff_ffffu32;
+
+const W: usize = std::mem::size_of::<usize>();
+const N: usize = 5;
+
+static CRC32_BYTE_TABLE: [[u32; 256]; 1] = build_crc32_table(1);
+static CRC32_WORD_TABLE: [[u32; 256]; W] = build_crc32_table(1);
+static CRC32_STRIDE_TABLE: [[u32; 256]; W] = build_crc32_table(N);
```
folkertdev, line 10: these could use some comments 

- you mentioned crc32 using different polynomials, maybe say something about the choice here (though I guess the png just says: we're using this one).
- what is `N`?
- how are these tables used

zlib-rs/src/crc32.rs
```diff
@@ -0,0 +1,133 @@
...
+        // is needed in case size_of(usize) < size_of(u32)
+        if std::mem::size_of_val(&word_crc) <= W {
+            word_crc
+        } else {
+            (crc >> 8) ^ word_crc
+        }
+    })
+}
+
+pub fn crc32_words(data: &[u8]) -> u32 {
```
folkertdev, line 65: is this strictly a performance optimization?

zlib-rs/src/crc32.rs
```diff
@@ -0,0 +1,133 @@
...
+            std::array::from_fn(|j| usize::to_le(words[i * N + j] ^ (crcs[j] as usize)));
+
+        for j in 0..N {
+            // TODO: this doesn't compile
+            // Move overflow bits into the "next" word.
+            //if std::mem::size_of_val(&crcs[i]) > W {
+            //    crcs[i + 1] ^= crcs[i] >> (8 * W);
+            //}
+            crcs[j] = 0;
+        }
```
folkertdev, line 97: I'm not sure what is or should be happening here. Assuming the current logic is correct, this can be written at `crcs[0..N].fill(0)`.

zlib-rs/src/crc32.rs
```diff
@@ -0,0 +1,133 @@
...
+    crc32_naive_inner(data, START_CHECKSUM) ^ FINAL_XOR
+}
+
+fn crc32_words_inner(words: &[usize], start: u32, per_word_crcs: &[u32]) -> u32 {
+    words.iter().enumerate().fold(start, |crc, (i, word)| {
+        let mut word_crc = 0u32;
+        let value = *word ^ (crc ^ per_word_crcs.get(i).unwrap_or(&0)) as usize;
+        for (i, b) in value.to_le_bytes().into_iter().enumerate() {
+            word_crc ^= CRC32_WORD_TABLE[i][usize::from(b)];
+        }
```
folkertdev, line 53: you can `zip` the `CRC32_WORD_TABLE` here

folkertdev (commented):
left some comments. 

For usage in the gzip logic it's important to have a function that can take in an existing checksum value and more bytes, and update the checksum with those bytes. It's not obvious how this implementation would do that to me (but hopefully that is simple to add?)

folkertdev:
also as a general observation this already seems bigger and more complex than what `crc32fast` does in the general case . e.g. 

https://github.com/srijs/rust-crc32fast/blob/master/src/baseline.rs

is quite simple to follow. Am I missing something that is implemented here but not there (besides the tables, which apparently should preferably be stored in a static, not a const https://github.com/srijs/rust-crc32fast/commit/e61ce6a39bbe9da495198a4037292ec299e8970f I did not know this, but that is useful information).

zlib-rs/src/crc32.rs
ahomescu: Right, that was something I wanted to have a discussion about and defer the decision to you. Theoretically a Rust `usize` can be as small as 16 bits (note that `usize` implements `From<u16>` but not `From<u32>`) and as large as 128 (RISC-V) or even 256 bits (CHERI capabilities). That being said, it's probably safe to assume it's at least 32 bits and I'm definitely in favor of adding an assertion.

zlib-rs/src/crc32.rs
ahomescu: I'm following the algorithm from https://github.com/zlib-ng/zlib-ng/blob/develop/doc/crc-doc.1.0.pdf, and they handle two cases here:
* The size of the CRC (`D` in the paper) is larger than the word size
* The CRC is smaller than the word size.

Since we use `u32` for the CRC and `usize` for the word, we could ignore the first case and assert that `u32 <= usize`.

zlib-rs/src/crc32.rs
ahomescu: It was also useful for testing during development, but it could go away now that the rest is done.

`crc32_words_inner` is still needed because `crc32_interleaved` calls it.

ahomescu:
> For usage in the gzip logic it's important to have a function that can take in an existing checksum value and more bytes, and update the checksum with those bytes. It's not obvious how this implementation would do that to me (but hopefully that is simple to add?)

Should be pretty simple, we can replace `START_CHECKSUM` at its uses with the given checksum value. I think even `crc32_interleaved` can support that.

ahomescu:
> also as a general observation this already seems bigger and more complex than what `crc32fast` does in the general case

That should be pretty simple to add by using the tables from `build_crc32_table`. My implementation here is based on zlib-ng's "braid" algorithm, i.e., the "interleaved" approach from https://github.com/zlib-ng/zlib-ng/blob/develop/doc/crc-doc.1.0.pdf. Is that overkill? Up to you, let me know how you'd like to proceed.

folkertdev:
In that case I think it makes more sense to follow the zlib-ng implementation more closely (https://github.com/zlib-ng/zlib-ng/blob/develop/arch/generic/crc32_braid_c.c). it seems to have diverged somewhat from the original paper. This is often my experience with CS literature: the way to write something up for an academic paper is generally not how you actually implement things. It will be interesting to see how this compares to the `crc32fast` baseline algorithm (I'd hope the zlib variant is faster, otherwise what are they doing, right?)

The implementation in that C file seems straightforward, but we can probably clean it up given that we're using rust. But I think it's a good idea to use the same name (braid), type signature, and generally follow its structure.

I think using an array for the `N + 1` crcs and other values is a clever idea. LLVM should have no problem with that. The braid function can take N as a `const N: usize>` const generic so it's clearer why we're doing things that way.

There is some arithmetic with `0xFFFF_FFFF` that we can simplify because we know a `u32` is 32 bits, (not a `c_int` that might be wider). so an `xor 0xFFFF_FFFF` is just a bitwise negation for us.

using macros to unroll without duplicating too much code is fine (our adler32 implementation does this too).

folkertdev:
could you add some `#[allow(unused)]` to make the clippy warnings go away? I'd do it myself but I think your git settings don't allow for it. 

---

I've been playing around with the simd version, and from what I see `zlib-ng` does much more manual work than it really needs. The `crc32fast` implementation is basically the same speed at a fraction of the code size.

```
Benchmark 1 (87 runs): cargo run --release --example crc32_bench sse silesia-small.tar
  wall_time          57.3ms ± 3.38ms    55.5ms … 84.6ms          7 ( 8%)        0%
  peak_rss           21.8MB ±  138KB    21.6MB … 22.2MB          0 ( 0%)        0%
  cpu_cycles          151M  ± 3.99M      147M  …  181M           7 ( 8%)        0%
  instructions        325M  ± 46.0K      325M  …  325M           2 ( 2%)        0%
  cache_references   3.70M  ± 49.5K     3.59M  … 3.83M           0 ( 0%)        0%
  cache_misses        980K  ±  109K      832K  … 1.36M           5 ( 6%)        0%
  branch_misses       746K  ± 30.6K      732K  … 1.02M           3 ( 3%)        0%
Benchmark 2 (88 runs): cargo run --release --example crc32_bench crc32fast silesia-small.tar
  wall_time          57.4ms ± 2.19ms    55.5ms … 69.7ms          3 ( 3%)          +  0.1% ±  1.5%
  peak_rss           21.8MB ±  156KB    21.6MB … 22.2MB          0 ( 0%)          +  0.1% ±  0.2%
  cpu_cycles          152M  ± 4.85M      147M  …  185M           2 ( 2%)          +  0.6% ±  0.9%
  instructions        323M  ± 46.4K      323M  …  323M           1 ( 1%)          -  0.6% ±  0.0%
  cache_references   3.71M  ±  100K     3.61M  … 4.40M           3 ( 3%)          +  0.2% ±  0.6%
  cache_misses        995K  ±  135K      849K  … 1.40M           6 ( 7%)          +  1.6% ±  3.7%
  branch_misses       746K  ± 14.4K      732K  …  823K           5 ( 6%)          +  0.1% ±  0.9%
```

---

But that benchmark measures performance when all the data is available. Zlib can en/decode in a streaming fashion, and the crc32 algorithm it uses is tuned for that: `crc32fast` only stores a u32 as its state, but zlib-ng actually keeps 4 simd values around to store the state. So when the data comes in in small increments, zlib-ng should be saving a bunch of work. It's hard to tell how much that matters in practice though. I'll try to run some benchmarks for that too

Also we do need the fused crc with memcpy, and I don't think that fits into the current `crc32fast` design. Somehow all of these bits need to be synthesized together.

With all that, if you want to make more open source contributions, I think you all have a better setup for running avx-512 SIMD code, so you might look into adding a 512-bit simd version based on https://github.com/zlib-ng/zlib-ng/blob/develop/arch/x86/crc32_fold_vpclmulqdq_tpl.h to crc32fast itself. Assuming our code will be close to what `crc32fast` does we can then use that for zlib-rs down the line.

folkertdev:
hah, wow

```
Benchmark 1 (84 runs): cargo run --release --example crc32_bench sse-chunked silesia-small.tar
  wall_time          59.3ms ± 1.77ms    57.0ms … 64.6ms          4 ( 5%)        0%
  peak_rss           21.8MB ±  156KB    21.5MB … 22.1MB          0 ( 0%)        0%
  cpu_cycles          158M  ± 3.51M      154M  …  174M           6 ( 7%)        0%
  instructions        354M  ± 49.4K      354M  …  354M           2 ( 2%)        0%
  cache_references   3.76M  ± 57.0K     3.67M  … 3.94M           1 ( 1%)        0%
  cache_misses       1.09M  ±  161K      880K  … 1.52M           3 ( 4%)        0%
  branch_misses       742K  ± 6.52K      731K  …  771K           1 ( 1%)        0%
Benchmark 2 (55 runs): cargo run --release --example crc32_bench crc32fast-chunked silesia-small.tar
  wall_time          90.8ms ± 3.19ms    87.9ms …  111ms          1 ( 2%)        💩+ 53.2% ±  1.4%
  peak_rss           21.8MB ±  151KB    21.5MB … 22.2MB          0 ( 0%)          +  0.1% ±  0.2%
  cpu_cycles          280M  ± 7.58M      274M  …  329M           1 ( 2%)        💩+ 77.1% ±  1.2%
  instructions        434M  ± 60.2K      434M  …  434M           3 ( 5%)        💩+ 22.6% ±  0.0%
  cache_references   3.78M  ±  115K     3.59M  … 4.48M           1 ( 2%)          +  0.6% ±  0.8%
  cache_misses       1.18M  ±  173K      934K  … 1.63M           0 ( 0%)        💩+  8.3% ±  5.2%
  branch_misses       747K  ± 14.8K      735K  …  831K           3 ( 5%)          +  0.6% ±  0.5%
```

that makes a _big_ difference! This is running

```
            let mut h = crc32fast::Hasher::new_with_initial(0);

            for c in input.chunks(32) {
                h.update(c);
            }
            println!("{:#x}", h.finalize());
```

and the equivalent thing for the zlib-ng approach. So, that is something that we might also consider upstreaming after we've cleaned it up considerably (the crc32fast code looks a lot nicer than the zlib-ng version)

folkertdev (approved):
thanks for implementing this!

I'll just fix the clippy issues on main.

ahomescu:
Most of the clippy issues should go away once another part of the code starts using these functions.

PR #28, opened 2024-02-09
Deflate codecov attempt 2

folkertdev:
trying to reach more branches in the tests


zlib-rs/src/deflate.rs
```diff
@@ -906,6 +906,7 @@ impl<'a> State<'a> {
             self.bi_buf |= val << self.bi_valid;
             self.bi_valid = total_bits;
         } else if self.bi_valid == Self::BIT_BUF_SIZE {
+            // with how send_bits is called, this is unreachable in practice
```
bramtweedegolf, line 909: That's interesting

bramtweedegolf (approved):

zlib-rs/src/deflate.rs
folkertdev: I'm guessing static analysis trips over the logic in the branch above noet being valid for high shifts (shift with 64 or higher is UB in C).

PR #33, opened 2024-02-19
Gzip support for inflate

bramtweedegolf:
Partial gzip implementation for inflate


zlib-rs/src/inflate.rs
```diff
@@ -450,10 +498,21 @@ const fn zswap32(q: u32) -> u32 {
 const INFLATE_FAST_MIN_HAVE: usize = 15;
 const INFLATE_FAST_MIN_LEFT: usize = 260;
 
+// only supported compression method
+const Z_DEFLATED: u64 = 8;
```
folkertdev, line 502: we should actually define this in `c_api.rs` and use it from there

zlib-rs/src/inflate.rs
```diff
@@ -489,16 +548,24 @@ impl<'a> State<'a> {
...
         need_bits!(self, 16);
 
-        // TODO gzip
+        // Gzip
+        if (self.wrap & 2) != 0 && self.bit_reader.hold() == 0x8b1f {
+            if self.wbits == 0 {
+                self.wbits = 15;
+            }
+
+            // TODO: CRC
```
folkertdev, line 557: so, this will happen later? seems important?

zlib-rs/src/inflate.rs
```diff
@@ -533,6 +600,252 @@ impl<'a> State<'a> {
...
+        self.extra()
+    }
+
+    fn extra(&mut self) -> ReturnCode {
+        if (self.flags & 0x0400) != 0 {
+
+            let mut copy = self.length;
+            if copy > self.in_available {
+                copy = self.in_available;
+            }
```
folkertdev, line 683:
```suggestion
            let copy = Ord::min(self.length, self.in_available);
```

zlib-rs/src/inflate.rs
```diff
@@ -533,6 +600,252 @@ impl<'a> State<'a> {
...
+
+                // Take a byte
+                // TODO: can this be done in a single line?
+                //
+                // This does not go well after a few pulls:
+                // let next_byte = self.bit_reader.pull_byte();
+                //
+                need_bits!(self, 8);
+                let len = self.bit_reader.hold() as u8;
+                self.bit_reader.init_bits();
```
folkertdev, line 748: what is the problem here?

zlib-rs/src/inflate.rs
```diff
@@ -541,12 +854,11 @@ impl<'a> State<'a> {
         if self.wrap != 0 {
             need_bits!(self, 32);
 
+            // TODO gzip
             if self.wrap & 4 != 0 && !self.writer.filled().is_empty() {
                 self.checksum = adler32(self.checksum, self.writer.filled());
             }
```
folkertdev, line 860: this needs to be fixed then, right?

zlib-rs/src/inflate.rs
```diff
@@ -1635,6 +1947,17 @@ pub unsafe fn copy(dest: *mut z_stream, source: &InflateStream) -> ReturnCode {
...
+    // Deep copy the Gzip header.
+    // TODO: is that what we want here?
+    let mut head_copy;
+    let head_opt = match &state.head {
+      Some(head) => {
+        head_copy = *(*head);
+        Some(&mut head_copy)
+      },
+      None => None,
+    };
```
folkertdev, line 1959: I don't think so. This is why this function is rather nasty: the `InflateStream` contains mutable references that we cannot just copy. That is why this `addr_of_mut!` stuff is needed in this function. We´ll need something similar here. 

we _cannot_  duplicate the reference, hence the copied value cannot be considered initialized.

zlib-rs/src/inflate.rs
```diff
@@ -1835,3 +2159,18 @@ fn init_window<'a>(
 
     Ok(window)
 }
+
+pub fn get_header<'a>(stream: &'a mut InflateStream<'a>, head: &'a mut GzipHeader) -> ReturnCode {
+
+    // TODO Why does this always fail?
+    //if (stream.state.wrap & 2) == 0 {
+    //    return ReturnCode::StreamError;
+    //}
```
folkertdev, line 2168: guess: you need a state that is initialized with a memory level in the range `16..32`

folkertdev (commented):
first pass

zlib-rs/src/c_api.rs
```diff
@@ -124,3 +124,38 @@ pub const Z_HUFFMAN_ONLY: c_int = 2;
...
+    /// true for a header CRC16
+    pub hcrc: c_int,
+
+    /// true when done reading header
+    pub done: c_int,
+}
+
+pub type gz_headerp = *mut gz_header;
+
+pub const INFLATE_STATE_SIZE: usize = core::mem::size_of::<crate::inflate::State>();
```
folkertdev, line 161: this constant is not really part of the C api, right? in any case we should preserve the comment that this should only be exposed during tests.

zlib-rs/src/inflate.rs
```diff
@@ -14,15 +15,34 @@ use crate::{
...
+    pub extra: *mut crate::c_api::Bytef,
+    pub extra_len: crate::c_api::uInt,
+    pub extra_max: crate::c_api::uInt,
+    pub name: *mut crate::c_api::Bytef,
+    pub name_max: crate::c_api::uInt,
+    pub comment: *mut crate::c_api::Bytef,
+    pub comment_max: crate::c_api::uInt,
+    pub hcrc: libc::c_int,
+    pub done: libc::c_int,
+}
```
folkertdev, line 43: this struct does not add any additional safety or guarantees, so I think we can just use `gz_header` everywhere. Just saves code.

zlib-rs/src/inflate.rs
```diff
@@ -533,6 +567,262 @@ impl<'a> State<'a> {
...
+                                    dest as usize
+                                );
+                            }
+                        }
+                    }
+                }
+
+                // Checksum
+                if (self.flags & 0x0200) != 0 && (self.wrap & 4) != 0 {
+                    let bytes = unsafe { slice::from_raw_parts(self.bit_reader.as_ptr(), copy) };
```
folkertdev, line 687: can this use `BitReader::as_slice()[..copy]`?

zlib-rs/src/inflate.rs
```diff
@@ -533,6 +567,262 @@ impl<'a> State<'a> {
...
+            head.name = std::ptr::null_mut();
+        }
+
+        self.length = 0;
+        self.mode = Mode::Comment;
+        self.comment()
+    }
+
+    fn comment(&mut self) -> ReturnCode {
+        assert!(self.length == 0);
```
folkertdev, line 755: use `assert_eq!`

folkertdev (commented):
please also apply `cargo fmt`

libz-rs-sys/src/tests/inflate.rs
```diff
@@ -182,7 +183,37 @@ fn inf(input: &[u8], _what: &str, step: usize, win: i32, len: usize, err: c_int)
...
+            extra: extra.as_ptr() as *mut u8,
+            extra_len: 0,
+            extra_max: 64,
+            name: name.as_ptr() as *mut u8,
+            name_max: 64, // How / where should this be set?
+            comment: comment.as_ptr() as *mut u8,
+            comment_max: 64,
+            hcrc: 0,
+            done: 0,
+        };
```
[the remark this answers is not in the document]
folkertdev: ah I get it now: those arrays need to be moved out of the `if` so that their lifetime actually matches their use

PR #43, opened 2024-02-21
Implemented need dict test

bramtweedegolf:
Fix: set_dictionary should return a DataError, not StreamError.
Also implemented the need_dictionary test.

### DIFF
--- a/zlib-rs/src/inflate.rs
+++ b/zlib-rs/src/inflate.rs
@@ -280,9 +280,9 @@ struct Table {
     bits: usize,
 }
 
-pub struct State<'a> {
+pub(crate) struct State<'a> {
     /// Current inflate mode
-    pub mode: Mode,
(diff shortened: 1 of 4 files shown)


zlib-rs/src/inflate.rs
```diff
@@ -280,9 +280,10 @@ struct Table {
     bits: usize,
 }
 
-pub(crate) struct State<'a> {
+pub struct State<'a> {
     /// Current inflate mode
-    mode: Mode,
+    pub mode: Mode,
+
```
folkertdev, line 286: hmm yeah we really don't want to make this public. The options to work around that are kind of limited though. I think it's best to make some `pub unsafe fn set_mode_dict(strm: &mut z_stream)` to inflate. We can hide that funtion in the future with feature flags.

PR #81, opened 2024-04-09
add feature flag to configure default allocator for libz-rs-sys

folkertdev:


libz-rs-sys/Cargo.toml
```diff
@@ -13,6 +13,11 @@ rust-version.workspace = true
 [lib]
 crate-type=["rlib", "cdylib"]
 
+[features]
+default = ["rust-allocator"] # when used as a rust crate, use the rust allocator
+c-allocator = [] # by default, use malloc/free for memory allocation
+rust-allocator = [] # by default, use the rust global alloctor for memory allocation
```
folkertdev, line 19: I think this is the right default. When we build a dynamic library from this crate we should probably use the `c-allocator` to be consistent with stock zlib.

bramtweedegolf: Agreed

rnijveld (approved):

PR #89, opened 2024-04-17
run CI on macos

folkertdev:


folkertdev:
it turns out that zlib-ng produces different output between x86 and aarch64. we produce the same output on both platforms (modulo the gzip header which contains a byte to inidicate the operating system). 

issue https://github.com/memorysafety/zlib-rs/issues/91

based on the other tests and fuzzer our implementation is correct on aarch64, so I will disable these tests on aarch64 for now.

rnijveld (approved):

PR #97, opened 2024-05-10
feat: no_std

JonasKruckenberg:
As promised:
This PR adds `no_std` support to `zlib-rs`, by adding two new feature flags `std` and `alloc` that hide related functionality behind them (enabled by default).

I also had to disable the auto-detection of features which is not yet exposed through feature flags.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,6 @@ libz-sys = { version = "1.1.12", default-features = false, features = ["zlib-ng"
 arbitrary = { version = "1.0" }
 quickcheck = { version = "1.0.3", default-features = false, features = [] }
 
-zlib-rs = { version = "0.1.1", path = "./zlib-rs" }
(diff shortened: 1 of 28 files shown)


folkertdev:
I made it so that `cargo check` at least works for `--target thumbv7em-none-eabihf` now with the `zlib-rs` crate itself. If you're OK with some API breakage in the future that should provide everything you need to play around with this I think. 

But maybe not, let me know. Also

- can you rebase this on main. I think I messed something up in git and I cannot force push to this branch
- do you have thoughts on how to test this in CI? would a working (but with warnings) `cargo check --target thumb*` provide enough stability?

folkertdev:
turns out I can force push. I think this is ready to merge, but would like to try it out on actual embedded hardware just to make sure there are no unexpected issues. I'll try to do that tomorrow

PR #110, opened 2024-06-07
fix bug in match copying

### LINKED ISSUE #109: large chunked output hits `todo!()` in inflate
reported on the flate2 repo https://github.com/rust-lang/flate2-rs/issues/413

folkertdev:
fix #109

The true cause is that in many calculations, the padding that is added to the window (so that SIMD operations don't need to perform bounds checks) was included in calculations where it should not be. This caused wrapping behavior in the window to be incorrect, and things go sideways from there.

Correcting this bug and subsequent fuzzing found  test cases that hit all the branches in the match copying logic, which is now fully implemented.

thanks to @VorpalBlade for the original bug report and easy reproduction of the problem!


folkertdev:
I used this fuzzer to find input that triggers the `op < len` case in the else branch of inffast
<details>
  <summary>Click me</summary>

```rust
#![no_main]
use libfuzzer_sys::fuzz_target;

use crate::arbitrary::{Arbitrary, Unstructured};
use core::ffi::c_int;
use core::mem::MaybeUninit;
use libfuzzer_sys::arbitrary;
use zlib_rs::{deflate::DeflateConfig, InflateFlush, ReturnCode};

#[derive(Debug)]
struct Input {
    level: i32,       // valid range 1..=9
    window_bits: i32, // valid range is -15..=-1
    mem_level: i32,   // valid range = 8..=9
    data: Vec<u8>,
    buf_size: usize, // valid range is 1..=(8 * 1024)
}
impl Arbitrary<'_> for Input {
    fn arbitrary(u: &mut Unstructured<'_>) -> arbitrary::Result<Self> {
        let level = u.int_in_range(1..=9)?;
        let window_bits = u.int_in_range(-15..=-9)?;
        let mem_level = u.int_in_range(8..=9)?;
        let buf_size = u.int_in_range(1..=8 * 1024)?;
        let data = Vec::<u8>::arbitrary(u)?;

        Ok(Input {
            level,
            window_bits,
            mem_level,
            data,
            buf_size,
        })
    }
}

fuzz_target!(|data: Input| {
    use libz_rs_sys::*;

    let Input {
        level,
        window_bits,
        mem_level,
        data,
        buf_size,
    } = data;

    if data.is_empty() {
        return;
    }

    let config = DeflateConfig {
        level,
        window_bits,
        mem_level,
        ..Default::default()
    };

    let mut input = vec![0; 1 << 17];
    let (input, err) = zlib_rs::deflate::compress_slice(&mut input, &data, config);
    assert_eq!(err, ReturnCode::Ok);

    // dbg!(level, window_bits, mem_level, buf_size,);

    //    std::fs::write(
    //        "/home/folkertdev/rust/zlib-rs/libz-rs-sys/src/tests/test-data/op-len-edge-case.zraw",
    //        &input,
    //    )
    //    .unwrap();

    let mut buf = vec![0; buf_size];

    {
        let mut stream = MaybeUninit::<z_stream>::zeroed();

        let err = unsafe {
            inflateInit2_(
                stream.as_mut_ptr(),
                window_bits,
                zlibVersion(),
                core::mem::size_of::<z_stream>() as c_int,
            )
        };
        assert_eq!(ReturnCode::from(err), ReturnCode::Ok);

        let stream = unsafe { stream.assume_init_mut() };

        stream.next_in = input.as_ptr() as *mut u8;
        stream.avail_in = input.len() as _;

        while stream.avail_in != 0 {
            stream.next_out = buf.as_mut_ptr();
            stream.avail_out = buf.len() as _;

            let err = unsafe { inflate(stream, InflateFlush::NoFlush as _) };

            if ReturnCode::from(err) == ReturnCode::BufError {
                stream.avail_out = buf.len() as _;
                continue;
            }
        }

        unsafe { inflateEnd(stream) };
    }
});
```
</details>

zlib-rs/src/inflate.rs
```diff
@@ -863,7 +863,7 @@ impl<'a> State<'a> {
 
     fn lit(&mut self) -> ReturnCode {
         if self.writer.remaining() == 0 {
-            #[cfg(feature = "std")]
+            #[cfg(all(test, feature = "std"))]
```
folkertdev, line 866: this and similar cases just limit what gets printed. in tests this could be useful information, in production you do not want this.

zlib-rs/src/inflate.rs
```diff
@@ -1242,11 +1242,20 @@ impl<'a> State<'a> {
...
+                wsize - copy
+            } else {
+                wnext - copy
+            };
+
+            copy = Ord::min(copy, self.length);
             copy = Ord::min(copy, left);
 
-            self.writer.extend(&slice[..copy]);
+            self.writer.extend(&self.window.as_slice()[from..][..copy]);
```
folkertdev, line 1258: mirrors the zlib-ng logic more exactly. the old logic was buggy after the window padding correction.

zlib-rs/src/inflate.rs
```diff
@@ -1592,22 +1603,46 @@ fn inflate_fast_help(state: &mut State, _start: usize) -> ReturnCode {
...
+                                if op < len as usize {
+                                    // This case is hit when part of the match is at the end of the
+                                    // window, and part of it has wrapped around to the start. Copy
+                                    // the end section here, the start section will be copied below.
+                                    len -= op as u16;
+                                    writer.extend(&state.window.as_slice()[from..][..op]);
+                                    from = 0;
+                                    op = window_next;
+                                }
                             }
```
folkertdev, line 1639: core achievement of this PR: we got rid of the `todo!`s here because we finally have test cases that hit all the branches!

zlib-rs/src/inflate/window.rs
```diff
@@ -20,24 +20,20 @@ pub struct Window<'a> {
...
+        self.size() == 0
     }
 
     pub fn size(&self) -> usize {
-        self.buf.len()
+        if self.buf.is_empty() {
+            0
+        } else {
+            self.buf.len() - Self::padding()
+        }
```
folkertdev: the true cause of these problems: padding should not be part of the window size! An empty buffer is only used in tests or when deallocating the window.

zlib-rs/src/inflate/window.rs
```diff
@@ -195,68 +184,82 @@ unsafe fn slice_assume_init(slice: &[MaybeUninit<u8>]) -> &[u8] {
...
+        window
+    }
+
     #[test]
     fn extend_in_bounds() {
         let mut checksum = 0;
 
-        let mut buf = [0; 12];
-        let mut window = Window::from_slice(&mut buf);
+        let mut window = init_window(4);
```
folkertdev, line 203: we now actually allocate the window, so these tests needed to be slightly adjusted because a window is now always some power of 2 bytes

libz-rs-sys/src/tests/inflate.rs
```diff
@@ -1446,3 +1446,255 @@ fn version_error() {
     };
     assert_eq!(ret, Z_VERSION_ERROR);
 }
+
+#[test]
+fn issue_109() {
```
folkertdev, line 1451: tests for  the various issues. not very exciting.

zlib-rs/src/inflate.rs
VorpalBlade: Thanks, that is useful, apart form the crash, these outputs were messing up the structured output of my command line program. A library should never have prints like this, instead consider using the log or tracing frameworks if the outputs do belong in production.

folkertdev: absolutely. It was unclear earlier to what extend these errors were fatal (and should be unreachable). We now know that they are part of the standard (de)compression workflow. So while in testing they may provide useful information, we don't want the prints in production for the reasons you mention.

rnijveld (approved):

PR #113, opened 2024-06-12
Fix inflate() flush modes and implement state.data_type

cjgriscom:
This PR address the following issues:

- `state.data_type` is always 0
- https://github.com/memorysafety/zlib-rs/issues/114
- Add logic to advance from `TYPE` state to `TYPE_DO` state

### DIFF
--- a/zlib-rs/src/inflate.rs
+++ b/zlib-rs/src/inflate.rs
@@ -252,6 +252,7 @@ pub enum Mode {
     Stored,
     CopyBlock,
     Check,
+    Len_,
     Len,
     Lit,
     LenExt,
@@ -505,6 +506,7 @@ impl<'a> State<'a> {
             Mode::CopyBlock => self.copy_block(),
             Mode::Check => self.check(),
             Mode::Len => self.len(),
+            Mode::Len_ => self.len_(),
             Mode::LenExt => self.len_ext(),
             Mode::Lit => self.lit(),
             Mode::Dist => self.dist(),
@@ -965,13 +967,14 @@ impl<'a> State<'a> {
                     bits: 5,
                 };
 
+                self.mode = Mode::Len_;
+
                 self.bit_reader.drop_bits(2);
 
                 if let InflateFlush::Trees = self.flush {
                     self.inflate_leave(ReturnCode::Ok)
                 } else {
-                    self.mode = Mode::Len;
-                    self.len()
+                    self.len_()
                 }
             }
             2 => {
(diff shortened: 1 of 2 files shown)


zlib-rs/src/inflate.rs
```diff
@@ -1922,32 +1927,38 @@ pub unsafe fn inflate(stream: &mut InflateStream, flush: InflateFlush) -> Return
...
         assert!(msg.ends_with(|c| c == '\0'));
         stream.msg = msg.as_ptr() as *mut u8 as *mut core::ffi::c_char;
     }
 
+    stream.data_type =
+        state.bit_reader.bits_in_buffer() as i32 +
+        if state.last { 64 } else { 0 } +
+        if matches!(state.mode, Mode::Type) { 128 } else {0} +
+        if matches!(state.mode, Mode::Len | Mode::CopyBlock) { 256 } else { 0 };
+
```
folkertdev, line 1961: because we're only ever in one mode, I think a `match` is nicer here. Also can tests be added that verify the behavior (probably from the libz-rs-sys` crate's test modules) versus zlib itself?

zlib-rs/src/inflate.rs
```diff
@@ -1880,6 +1880,11 @@ pub unsafe fn inflate(stream: &mut InflateStream, flush: InflateFlush) -> Return
     let dest_slice = core::slice::from_raw_parts_mut(stream.next_out, stream.avail_out as usize);
 
     let state = &mut stream.state;
+    if matches!(state.mode, Mode::Type) {
+        state.mode = Mode::TypeDo;
+    }
```
folkertdev, line 1885: can a test be constructed that verifies this behavior?

folkertdev (commented):
nice work! 

it's interesting that none of the zlib/zlib-ng tests catch these bugs. We'd like to add additional tests when we find that incorrect behavior slipped through.

zlib-rs/src/inflate.rs
cjgriscom: Hmm, you're right.  I will double-check zlib's original logic and correct it to one match.  I will also put some tests together that demonstrate parity b/t libz-rs-sys and libz-sys

cjgriscom:
I split the changes into 3 commits and added a test which compares the stream positions and state data after each call to inflate().  It demonstrates that all three changes are needed for consistency

folkertdev:
After your changes, other tests were failing. Turns out this is what the `LEN_` state is for. It looks really weird otherwise because it does no real work, just falls throught, but apparently that state is important for the correct behavior of `Flush::Trees`.

Also for the future, our CI checks formatting and runs `cargo clippy` (also on tests). We advise you run those tools locally at every commit, that'll save a bunch of back-and-forth with CI.

folkertdev (approved):
Thanks for these fixes!

cjgriscom:
Thanks for merging and for your work on this crate!

PR #118, opened 2024-06-21
include LICENSE file in published crates

decathorpe:
The Zlib license requires that redistributed sources contain a copy of the license text. This is currently not the case for zlib-rs and libz-rs-sys crates as published and redistributed via crates.io.

This PR adds symbolic links that point to the LICENSE file at the project root to both crates. This causes cargo to include the file when running `cargo publish` or `cargo package`.

Note that this might not work correctly when running `cargo package` or `cargo publish` on an operating system that does not support symbolic links (Windows).


folkertdev:
thanks!

PR #125, opened 2024-07-22
add `custom-prefix` feature flag

### LINKED ISSUE #124: libz-rs-sys functions are missing #[no_mangle]
```
$ cat > foo.rs <<EOF
#[no_mangle]
pub extern "C" fn foo() {}

pub extern "C" fn bar() {}
EOF
$ rustc --crate-type lib foo.rs
$ nm libfoo.rlib
nm libfoo.rlib 

lib.rmeta:
nm: lib.rmeta: no symbols

foo.foo.a77a36e9eb12c384-cgu.0.rcgu.o:
0000000000000000 T _ZN3foo3bar17h58f86806802c8f61E
0000000000000000 T foo
```

See how the symbol for functions is mangled unless `#[no_mangle]` is applied to it.

Without `[no_mangle]`, the library cannot be used as a replacement for libz for C code.

That said, in some cases, it is also desirable to rename the symbols completely. For instance, in the Firefox code base, [this is in place](https://searchfox.org/mozilla-central/rev/b220e40ff2ee3d10ce68e07d8a8a577d5558e2a2/modules/zlib/src/mozzconf.h#16-172) to avoid unexpectedly ending up using the system libz through some other system library.

folkertdev:
fixes #124

### DIFF
--- a/libz-rs-sys/Cargo.toml
+++ b/libz-rs-sys/Cargo.toml
@@ -18,6 +18,7 @@ default = ["std", "rust-allocator"] # when used as a rust crate, use the rust al
 c-allocator = ["zlib-rs/c-allocator"] # by default, use malloc/free for memory allocation
 rust-allocator = ["zlib-rs/rust-allocator"] # by default, use the rust global alloctor for memory allocation
 std = ["zlib-rs/std"] # assume `::std` is available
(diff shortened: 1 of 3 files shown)


rnijveld (approved):

rnijveld:
Looks good to me!

PR #150, opened 2024-08-16
test edge cases in the public C api

folkertdev:
we now cover all branches with dedicated tests.

### DIFF
--- a/zlib-rs/src/inflate.rs
+++ b/zlib-rs/src/inflate.rs
@@ -2221,14 +2221,11 @@ pub fn set_dictionary(stream: &mut InflateStream, dictionary: &[u8]) -> ReturnCo
     ReturnCode::Ok
 }
 
-pub fn end<'a>(stream: &'a mut InflateStream) -> &'a mut z_stream {
-    let mut state = State::new(&[], ReadBuf::new(&mut []));
-    core::mem::swap(&mut state, stream.state);
+pub fn end<'a>(stream: &'a mut InflateStream<'a>) -> &'a mut z_stream {
+    let alloc = stream.alloc;
 
     let mut window = Window::empty();
-    core::mem::swap(&mut window, &mut state.window);
-
-    let alloc = stream.alloc;
+    core::mem::swap(&mut window, &mut stream.state.window);
 
     // safety: window is not used again
     if !window.is_empty() {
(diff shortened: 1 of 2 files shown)


zlib-rs/src/inflate.rs
```diff
@@ -2221,14 +2221,11 @@ pub fn set_dictionary(stream: &mut InflateStream, dictionary: &[u8]) -> ReturnCo
     ReturnCode::Ok
 }
 
-pub fn end<'a>(stream: &'a mut InflateStream) -> &'a mut z_stream {
-    let mut state = State::new(&[], ReadBuf::new(&mut []));
-    core::mem::swap(&mut state, stream.state);
```
folkertdev, line 2226: miri reported UB here. I think the problem was that `state` got dropped and/or got confused about lifetimes here?

```
error: Undefined Behavior: constructing invalid value at .writer.buf: encountered a dangling reference (use-after-free)
    --> /home/folkertdev/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/mem/mod.rs:734:14
     |
734  |     unsafe { intrinsics::typed_swap(x, y) }
     |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ constructing invalid value at .writer.buf: encountered a dangling reference (use-after-free)
     |
     = note: BACKTRACE on thread `coverage::ub`:
     = note: inside `std::mem::swap::<zlib_rs::inflate::State<'_>>` at /home/folkertdev/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/mem/mod.rs:734:14: 734:42
note: inside `zlib_rs::inflate::end`
    --> /home/folkertdev/rust/zlib-rs/zlib-rs/src/inflate.rs:2226:5
     |
2226 |     core::mem::swap(&mut state, stream.state);
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside `libz_rs_sys::inflateEnd`
    --> /home/folkertdev/rust/zlib-rs/libz-rs-sys/src/lib.rs:370:13
     |
370  |             zlib_rs::inflate::end(stream);
```

but I'm not 100% sure. This fix makes miri happy, and still passes all tests so I think we're good.

PR #156, opened 2024-08-18
Rework inflate init

folkertdev:


folkertdev:
too much hassle for kinda no gain

PR #157, opened 2024-08-21
refinements to unsafe code

folkertdev:
among other things

- the `Init` and `GetHeader` functions now assume that their arguments are properly initialized in the rust sense. Providing uninitialized memory to these functions is UB, though unlikely to actually cause issues across an FFI boundary
-

### DIFF
--- a/libz-rs-sys/src/lib.rs
+++ b/libz-rs-sys/src/lib.rs
@@ -20,6 +20,7 @@ use core::mem::MaybeUninit;
 
 use core::ffi::{c_char, c_int, c_long, c_uchar, c_uint, c_ulong, c_void};
 
+use zlib_rs::allocate::Allocator;
 use zlib_rs::{
     deflate::{DeflateConfig, DeflateStream, Method, Strategy},
     inflate::{InflateConfig, InflateStream},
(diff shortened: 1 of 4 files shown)


libz-rs-sys/src/lib.rs
```diff
@@ -56,31 +65,6 @@ macro_rules! prefix {
...
-#[allow(unreachable_code)]
-const DEFAULT_ZFREE: Option<free_func> = '_blk: {
-    #[cfg(feature = "c-allocator")]
-    break '_blk Some(zlib_rs::allocate::Allocator::C.zfree);
-
-    #[cfg(feature = "rust-allocator")]
-    break '_blk Some(zlib_rs::allocate::Allocator::RUST.zfree);
-
-    None
-};
```
folkertdev, line 82: this is handled by zlib-rs. Only one allocator is available there because of the compile_error on line 66

libz-rs-sys/src/lib.rs
```diff
@@ -248,7 +226,9 @@ pub extern "C" fn adler32_combine(adler1: c_ulong, adler2: c_ulong, len2: z_off_
 ///
 /// The caller must guarantee that
 ///
-/// * The `destLen` pointer satisfies the requirements of [`core::ptr::read`]
+/// * Either
+///     - `destLen` is `NULL`
+///     - `destLen` satisfies the requirements of `&mut *destLen`
```
folkertdev, line 231: stock zlib happily dereferences a null pointer here. We check for it being null and return an error, which, because zlib has undefined behavior in this case, is compatible.

test-libz-rs-sys/src/lib.rs
```diff
@@ -409,7 +409,7 @@ mod null {
             let mut strm = MaybeUninit::<z_stream>::zeroed();
 
             fn initialize_header_null() -> MaybeUninit<gz_header> {
-                let mut head = MaybeUninit::<gz_header>::uninit();
+                let mut head = MaybeUninit::<gz_header>::zeroed();
```
folkertdev, line 413: this test uses some low-level operations still, but we now assume that the memory is initialized, so we have to use `zeroed` here.

PR #170, opened 2024-08-29
fix oversight in one of the quick deflate macros

### LINKED ISSUE #169: Z_BEST_SPEED doesn't handle Z_FINISH properly in some corner cases when there is not enough avail_out
Consider the following code:
```
#include <zlib.h>
#include <assert.h>
#include <string.h>
#include <stdio.h>

int main() {
    Bytef buf[4096];
    for (int i = 0; i < sizeof(buf); i++) {
        buf[i] = (i / 4) & 0xff;
    }
    Bytef out[4096];
    z_stream zs;
    memset(&zs, 0, sizeof(zs));
    int ret = deflateInit(&zs, Z_BEST_SPEED);
    assert(ret == Z_OK);
    zs.avail_in = 2048;
    zs.avail_out = 2048;
    zs.next_in = buf;
    zs.next_out = out;
    ret = deflate(&zs, Z_NO_FLUSH);
    assert(ret == Z_OK);

    zs.avail_in = 2048;
    zs.avail_out = 9;
    ret = deflate(&zs, Z_FINISH);
    assert(zs.avail_in == 0);
    return 0;
}
```
This works properly with zlib, but fails on the last assert with libz-rs-sys because zs.avail_in is still 2048.

Now, in this specific case, it still works afterwards (removing the last assert):
```
    zs.avail_out = sizeof(out) - zs.total_out;
    ret = deflate(&zs, Z_FINISH);
    assert(ret == Z_STREAM_END);
    assert(zs.avail_in == 0);
```

But depending on the input, you can end up with ret at Z_BUF_ERROR because it doesn't want avail_in to be non-zero, hitting the following code: https://github.com/memorysafety/zlib-rs/blob/e83fcefadb576e0ce32297cc516facb694d93b67/zlib-rs/src/deflate.rs#L2412-L2417

I haven't looked exactly why this isn't happening in the case above, but here's a testcase where it does happen:
<details>
<summary>testcase</summary>

```
include <zlib.h>
#include <assert.h>
#include <string.h>
#include <stdio.h>
#include <sys/mman.h>
#include <fcntl.h>

int main() {
    int fd = open("file.js", O_RDONLY);
    off_t len = lseek(fd, 0, SEEK_END);
    Bytef *buf = mmap(NULL, len, PROT_READ, MAP_PRIVATE, fd, 0);
    Bytef out[4096];
    z_stream zs;
    memset(&zs, 0, sizeof(zs));
    int ret = deflateInit2(&zs, Z_BEST_SPEED, Z_DEFLATED, -15, 8, Z_DEFAULT_STRATEGY);
    assert(ret == Z_OK);
    zs.avail_in = 2048;
    zs.avail_out = 1053;
    zs.next_in = buf;
    zs.next_out = out;
    ret = deflate(&zs, Z_NO_FLUSH);
    assert(ret == Z_OK);

    zs.avail_in = 67;
    ret = deflate(&zs, Z_FINISH);

    zs.avail_out = sizeof(out) - zs.total_out;
    ret = deflate(&zs, Z_FINISH);
    assert(ret == Z_STREAM_END);
    assert(zs.avail_in == 0);
    return 0;
}
```
file.js:
```
// This file was procedurally generated from the following sources:
// - src/dstr-binding/ary-init-iter-get-err.case
// - src/dstr-binding/error/cls-expr-async-gen-meth-static.template
/*---
description: Abrupt completion returned by GetIterator (static class expression async generator method)
esid: sec-class-definitions-runtime-semantics-evaluation
features: [Symbol.iterator, async-iteration]
flags: [generated]
info: |
    ClassExpression : class BindingIdentifieropt ClassTail

    1. If BindingIdentifieropt is not present, let className be undefined.
    2. Else, let className be StringValue of BindingIdentifier.
    3. Let value be the result of ClassDefinitionEvaluation of ClassTail
       with argument className.
    [...]

    14.5.14 Runtime Semantics: ClassDefinitionEvaluation

    21. For each ClassElement m in order from methods
        a. If IsStatic of m is false, then
        b. Else,
           Let status be the result of performing PropertyDefinitionEvaluation
           for m with arguments F and false.
    [...]

    Runtime Semantics: PropertyDefinitionEvaluation

    AsyncGeneratorMethod :
        async [no LineTerminator here] * PropertyName ( UniqueFormalParameters )
            { AsyncGeneratorBody }

    1. Let propKey be the result of evaluating PropertyName.
    2. ReturnIfAbrupt(propKey).
    3. If the function code for this AsyncGeneratorMethod is strict mode code, let strict be true.
       Otherwise let strict be false.
    4. Let scope be the running execution context's LexicalEnvironment.
    5. Let closure be ! AsyncGeneratorFunctionCreate(Method, UniqueFormalParameters,
       AsyncGeneratorBody, scope, strict).
    [...]

    13.3.3.5 Runtime Semantics: BindingInitialization

    BindingPattern : ArrayBindingPattern

    1. Let iterator be GetIterator(value).
    2. ReturnIfAbrupt(iterator).

---*/
var iter = {};
iter[Symbol.iterator] = function() {
  throw new Test262Error();
};


var C = class {
  static async *method([x]) {
    
  }
};

var method = C.method;

assert.throws(Test262Error, function() {
  method(iter);
});

reportCompare(0, 0);
```
</details>
Note: any other value of avail_out makes it not happen. Note that this value was picked because it's half the size of file.js.

folkertdev:
fix #169

### DIFF
--- a/zlib-rs/src/deflate/algorithm/quick.rs
+++ b/zlib-rs/src/deflate/algorithm/quick.rs
@@ -11,16 +11,6 @@ use crate::{
 pub fn deflate_quick(stream: &mut DeflateStream, flush: DeflateFlush) -> BlockState {
     let mut state = &mut stream.state;
 
-    let last = matches!(flush, DeflateFlush::Finish);
-
-    macro_rules! quick_start_block {
-        () => {
-            state.bit_writer.emit_tree(BlockType::StaticTrees, last);
-            state.block_open = 1 + last as u8;
-            state.block_start = state.strstart as isize;
-        };
-    }
-
     macro_rules! quick_end_block {
         ($last:expr) => {
             if state.block_open > 0 {
@@ -35,7 +25,7 @@ pub fn deflate_quick(stream: &mut DeflateStream, flush: DeflateFlush) -> BlockSt
                     state = &mut stream.state;
                 }
                 if stream.avail_out == 0 {
-                    return match last {
+                    return match $last {
                         true => BlockState::FinishStarted,
                         false => BlockState::NeedMore,
                     };
@@ -44,15 +34,25 @@ pub fn deflate_quick(stream: &mut DeflateStream, flush: DeflateFlush) -> BlockSt
         };
     }
 
+    macro_rules! quick_start_block {
+        ($last:expr) => {
+            state.bit_writer.emit_tree(BlockType::StaticTrees, $last);
+            state.block_open = 1 + $last as u8;
+            state.block_start = state.strstart as isize;
+        };
+    }
+
+    let last = matches!(flush, DeflateFlush::Finish);
+
     if last && state.block_open != 2 {
         /* Emit end of previous block */
         quick_end_block!(false);
         /* Emit start of last block */
-        quick_start_block!();
+        quick_start_block!(last);
     } else if state.block_open == 0 && state.lookahead > 0 {
         /* Start new block only when we have lookahead data, so that if no
         input data is given an empty block will not be written */
-        quick_start_block!();
+        quick_start_block!(last);
     }
 
     loop {
@@ -88,7 +88,7 @@ pub fn deflate_quick(stream: &mut DeflateStream, flush: DeflateFlush) -> BlockSt
             if state.block_open == 0 {
                 // Start new block when we have lookahead data,
                 // so that if no input data is given an empty block will not be written
-                quick_start_block!();
+                quick_start_block!(last);
             }
         }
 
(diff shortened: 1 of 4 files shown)


zlib-rs/src/deflate/algorithm/quick.rs
```diff
@@ -35,7 +25,7 @@ pub fn deflate_quick(stream: &mut DeflateStream, flush: DeflateFlush) -> BlockSt
                     state = &mut stream.state;
                 }
                 if stream.avail_out == 0 {
-                    return match last {
+                    return match $last {
```
folkertdev, line 28: this was the actual issue, `$last` was captured instead of the macro argument being used. I've moved the macros above the definition of `last` now so that this cannot happen by accident again.

PR #171, opened 2024-08-29
Use the "C-unwind" abi for functions we cannot guarantee won't panic

folkertdev:


folkertdev:
a note: `const extern "C-unwind"` is currently unstable https://github.com/rust-lang/rust/issues/64926, but our const functions should not panic anyway so having them be `extern "C"` should be fine

PR #173, opened 2024-08-30
fix issue 172

### LINKED ISSUE #172: Fails to decompress gzip streams at some chunk sizes
The following testcase, derived from https://github.com/web-platform-tests/wpt/blob/master/compression/decompression-split-chunk.tentative.any.js , fails with libz-rs-sys:

```
#include <zlib.h>
#include <assert.h>
#include <string.h>
#include <stdio.h>

int main(int argc, char*argv[]) {
    int exitcode = 0;
    Bytef buf[] = { 31, 139, 8, 0, 0, 0, 0, 0, 0, 3, 75, 173, 40, 72, 77, 46, 73, 77, 81, 200, 47, 45, 41, 40, 45, 1, 0, 176, 1, 57, 179, 15, 0, 0, 0 };
    for (uInt chunk = 1; chunk <= sizeof(buf); chunk++) {
        Bytef out[1024];
        z_stream zs;
        memset(&zs, 0, sizeof(zs));
        int ret = inflateInit2(&zs, 31);
        assert(ret == Z_OK);
        zs.avail_out = sizeof(out);
        zs.next_in = buf;
        zs.next_out = out;
        while (ret == Z_OK && zs.total_in < sizeof(buf)) {
            zs.avail_in = chunk > sizeof(buf) - zs.total_in ? sizeof(buf) - zs.total_in : chunk;
            ret = inflate(&zs, Z_NO_FLUSH);
        }
        if (ret != Z_STREAM_END) {
            fprintf(stderr, "Finished with %d at chunk size %u\n", ret, chunk);
            exitcode = 1;
        }
        if (memcmp(out, "expected output", zs.total_out) != 0) {
            fprintf(stderr, "Output did not match at chunk size %u\n", chunk);
            exitcode = 1;
        }
    }
    return exitcode;
}
```

It prints:
```
Finished with -3 at chunk size 8
Finished with -3 at chunk size 11
Finished with -3 at chunk size 16
Finished with -3 at chunk size 17
Finished with -3 at chunk size 31
Finished with -3 at chunk size 32
Finished with -3 at chunk size 33
Finished with -3 at chunk size 34
```

Note; the output check is just an extra check, it doesn't exhibit anything going wrong.

folkertdev:
fix #172

### DIFF
--- a/test-libz-rs-sys/src/inflate.rs
+++ b/test-libz-rs-sys/src/inflate.rs
@@ -1,4 +1,4 @@
-use core::ffi::{c_char, c_int, c_ulong, c_void, CStr};
+use core::ffi::{c_char, c_int, c_uint, c_ulong, c_void, CStr};
 use core::mem::{ManuallyDrop, MaybeUninit};
 
 use libz_rs_sys::*;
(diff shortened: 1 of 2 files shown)


test-libz-rs-sys/src/inflate.rs
```diff
@@ -1808,3 +1824,52 @@ fn test_inflate_flush_block() {
...
+                    chunk as c_uint
+                };
+                ret = inflate(strm, Z_NO_FLUSH);
+            }
+            if ret != Z_STREAM_END {
+                eprintln!("Finished with {} at chunk size {}\n", ret, chunk);
+                exitcode = 1;
+            }
+
+            if out[..strm.total_out as usize] != b"expected output"[..strm.total_out as usize] {
```
glandium, line 1865: The "expected output" should not be truncated.

folkertdev: right, cleaned that up in https://github.com/memorysafety/zlib-rs/pull/175

PR #176, opened 2024-08-31
Add the metadata needed by cargo-c

### LINKED ISSUE #174: Please use cargo-c to produce the cdylib
I can provide you with a PR, cdylibs do require some arch specific logic and cargo-c streamlines the whole process and provides also a pkg-config file and a platform-correct way to install the package.

lu-zero:
Fixes #174 

It does not install any header, uses `libz_rs` as name in pkg-config and `z_rs` as library name.

The default cargo `cinstall --destdir=/some/place`  would produce this on macos:

``` tree
└── usr
    └── local
        └── lib
            ├── libz_rs.0.2.1.dylib
            ├── libz_rs.0.2.dylib -> libz_rs.0.2.1.dylib
            ├── libz_rs.a
            ├── libz_rs.dylib -> libz_rs.0.2.1.dylib
            └── pkgconfig
                └── libz_rs.pc
```

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -325,7 +325,7 @@ jobs:
           targets: ${{matrix.target}}
       - name: Install cargo-c
         env:
-          LINK: https://github.com/lu-zero/cargo-c/releases/tag/v0.10.5/download
+          LINK: https://github.com/lu-zero/cargo-c/releases/download/v0.10.5
         run: |
           curl -L "$LINK/cargo-c-x86_64-unknown-linux-musl.tar.gz" |
           tar xz -C $HOME/.cargo/bin
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -325,7 +325,7 @@ jobs:
           targets: ${{matrix.target}}
       - name: Install cargo-c
         env:
-          LINK: https://github.com/lu-zero/cargo-c/releases/latest/download
+          LINK: https://github.com/lu-zero/cargo-c/releases/tag/v0.10.5/download
         run: |
           curl -L "$LINK/cargo-c-x86_64-unknown-linux-musl.tar.gz" |
           tar xz -C $HOME/.cargo/bin
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -251,7 +251,7 @@ jobs:
           done
 
   link-c-dynamic-library:
-    name: dynamic library
+    name: vanilla dynamic library
     strategy:
       matrix:
         include:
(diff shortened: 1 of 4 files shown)


libz-rs-sys-cdylib/README.md
```diff
@@ -74,6 +76,39 @@ the standard library has the following limitations:
...
+> tree target
+target
+├── CACHEDIR.TAG
+└── x86_64-unknown-linux-gnu
+    └── release
+        ├── libz_rs.a
+        ├── libz_rs.d
+        ├── libz_rs.pc
+        ├── libz_rs.so
+        └── libz_rs-uninstalled.pc
```
folkertdev, line 113: @lu-zero is there a reason `cargo cbuild` puts its things in `target/<target tripple>/release` rather than just `target/release`? it's annoyingly inconsistent with a standard `cargo build`.

lu-zero: It is required until cargo learns to differentiate the RUSTFLAGS for the build.rs and the RUSTFLAGS for the host.

using `cargo cbuild -v` it emits a note with the uninstalled pkg-config path.

lu-zero: https://github.com/rust-lang/cargo/issues/4423 is another way to see the problem and https://github.com/rust-lang/cargo/issues/9452 is how to solve it.

But normally `cinstall` is used to produce the `.a` that is then linked in the bigger C project. Even C projects support cross compilation ^^.

I guess I should really prepare a blogpost showing how to use cargo-c in many deployment scenario :)

.github/workflows/checks.yaml
```diff
@@ -304,6 +304,45 @@ jobs:
...
+        with:
+          persist-credentials: false
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@be73d7920c329f220ce78e0234b8f96b7ae60248
+        with:
+          toolchain: stable
+          targets: ${{matrix.target}}
+      - name: Install cargo-c
+        env:
+          LINK: https://github.com/lu-zero/cargo-c/releases/latest/download
```
folkertdev, line 328:
```suggestion
          LINK: https://github.com/lu-zero/cargo-c/releases/tag/v0.10.5/download
```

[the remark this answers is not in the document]
folkertdev: yeah, fixed that. though we may run into `cargo-c` and `cargo` inconsistencies as their versions drift, so this might need to be bumped periodically.

.github/workflows/checks.yaml
```diff
@@ -304,6 +304,45 @@ jobs:
...
+          curl -L "$LINK/cargo-c-x86_64-unknown-linux-musl.tar.gz" |
+          tar xz -C $HOME/.cargo/bin
+      - name: build with and test the result of cargo-c
+        working-directory: libz-rs-sys-cdylib
+        run: |
+          # build using cargo-c this time
+          cargo cinstall --offline --release --destdir=/tmp/cargo-cbuild-libz-rs
+          tree /tmp/cargo-cbuild-libz-rs
+          # verify that the SONAME  is set and includes a version
+          objdump -p target/x86_64-unknown-linux-gnu/release/libz_rs.so | awk '/SONAME/{print $2}' | grep -E 'libz_rs\.so\.[0-9]+\.[0-9]+'
```
bjorn3, line 339: The original zlib uses `libz.so.1` as SONAME. I think using `libz_rs.so.0.4` as SONAME for zlib-rs is wrong. We are matching the ABI of the original zlib and when zlib-rs bumps to version 0.5, the ABI will stay the same and thus SONAME shouldn't change.

.github/workflows/checks.yaml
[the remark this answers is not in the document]
lu-zero: I try to make a new `cargo-c` release the weekend the new `cargo` is released, so if you track the latest `cargo` you can track the latest `cargo-c` w/out, hopefully, issues.

.github/workflows/checks.yaml
lu-zero: you may set `package.metadata.capi.library.version` to address it.

.github/workflows/checks.yaml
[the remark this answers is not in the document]
folkertdev: would pinning the version always lead to issues, or does it usually work assuming no major changes to the parts of cargo that cargo-c uses?

libz-rs-sys-cdylib/Cargo.toml
```diff
@@ -19,6 +19,17 @@ default = ["c-allocator"] # when used as a cdylib crate, use the c allocator
 c-allocator = ["libz-rs-sys/c-allocator"] # by default, use malloc/free for memory allocation
 rust-allocator = ["libz-rs-sys/rust-allocator", "libz-rs-sys/std"] # by default, use the rust global alloctor for memory allocation
 custom-prefix = ["libz-rs-sys/custom-prefix"] # use the LIBZ_RS_SYS_PREFIX to prefix all exported symbols
+capi = []
 
 [dependencies]
 libz-rs-sys = { version = "0.4.0", path = "../libz-rs-sys", default-features = false }
+
+[package.metadata.capi.library]
```
lu-zero, line 27:
```suggestion
[package.metadata.capi.library]
version = "1.0.0"
```

.github/workflows/checks.yaml
folkertdev: hmm. That `0.4` is the rust version, and not the version of zlib that we mirror. We're not at 1.0 yet with zlib-rs, so in the meantime, how should we handle the situation?

I think mostly this c dylib is for individuals that want to use zlib-rs in their projects; we're not ready for distributions yet (I expect that to only happen at 1.0 and later). So is there a problem in practice here?

bjorn3: Using a version that doesn't have a relation to the ABI in SONAME kinda invalidates the point of having SONAME in the first place I would think. There are two main users of dylibs:

* Distros: These want SONAME to match the ABI compatibility such that they can upgrade the dylib without needing to rebuild everything when the new version is ABI compatible with the old version.
* Projects who bundle the dylib as part of their program: These don't need SONAME at all as the dylib will never be updated independently of the program that uses it.

libz-rs-sys-cdylib/Cargo.toml
folkertdev: nice, thanks!

.github/workflows/checks.yaml
folkertdev: cool, set that to `1.3.0` now, which is the actual api version that we say we provide (though some stuff is missing currently)

.github/workflows/checks.yaml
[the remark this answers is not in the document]
lu-zero: In general it should not, you'd end up in the same situation as if you were using the stable cargo-N with rustc-M.

But if you always use the latest `rustc` in the CI using the latest cargo-c should always work fine.

bjorn3:
I think it looks fine now, but I'm not really familiar with cargo-c.

folkertdev:
In the absence of cargo being able to generate a proper dynamic library, it's the best solution we have.

folkertdev (approved):
thanks @lu-zero for the help here!

PR #188, opened 2024-09-19
add specific tests for `copy_match`

folkertdev:


zlib-rs/src/inflate/writer.rs
```diff
@@ -176,12 +176,9 @@ impl<'a> Writer<'a> {
...
     ) {
         if (end - start).next_multiple_of(core::mem::size_of::<C>()) <= (buf.len() - current) {
+            let ptr = buf.as_mut_ptr();
             unsafe {
-                Self::copy_chunk_unchecked::<C>(
-                    buf.as_ptr().add(start),
-                    buf.as_mut_ptr().add(current),
-                    buf.as_ptr().add(end),
-                )
+                Self::copy_chunk_unchecked::<C>(ptr.add(start), ptr.add(current), ptr.add(end))
```
folkertdev, line 181: it is against miri rules to use the `as_ptr` to read from something modified by the `as_mut_ptr`. This is easily fixed by basing all pointers off of the mutable pointer. Because of subtyping, the `mut` is degrated to `const` when the type signatures demand it.

PR #190, opened 2024-09-19
Copy match simd

folkertdev:
the implementation is beautiful, but it does not actually help with our benchmarks. Overall `copy_match` is 1% to 2% of total runtime, so even though this should be faster, it really doesn't matter in the big picture.


folkertdev:
idk, maybe this will be useful at some point, but for now we just cannot justify that much unsafe for no measurable gain.

PR #203, opened 2024-09-26
fix UB in `inflate::State::extra`

folkertdev:
fixes https://github.com/memorysafety/zlib-rs/issues/195

I think this is a bit more elegant than a check for `count > 0`. 

Extra is a bit different because it has both a `extra_len` and `extra_max` field, so using the name/comment logic does not work.

cc @inahga

### DIFF
--- a/test-libz-rs-sys/src/inflate.rs
+++ b/test-libz-rs-sys/src/inflate.rs
@@ -1206,6 +1206,21 @@ fn gzip_chunked_2_bytes() {
 }
 
 #[test]
+fn gzip_chunked_15_bytes() {
+    gzip_chunked(15);
+}
+
+#[test]
+fn gzip_chunked_16_bytes() {
+    gzip_chunked(16);
+}
+
+#[test]
+fn gzip_chunked_17_bytes() {
+    gzip_chunked(17);
+}
+
+#[test]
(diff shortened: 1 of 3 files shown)


test-libz-rs-sys/src/inflate.rs
```diff
@@ -1205,6 +1205,11 @@ fn gzip_chunked_2_bytes() {
     gzip_chunked(2);
 }
 
+#[test]
+fn gzip_chunked_16_bytes() {
+    gzip_chunked(16);
+}
```
inahga, line 1211: Optional: you may consider throwing on in some non-powers-of-two checks as well, since there's no guarantee (AFAICT) a user will choose chunks that are powers of two. Or could throw `gzip_chunked()` into a `quickcheck`. 

(Then again, it looks like we do fuzz chunked inflate, so maybe it's not necessary.)

inahga (approved):
LGTM, thanks!

test-libz-rs-sys/src/inflate.rs
folkertdev: added some extra tests, but they are all good. 

We do fuzz chunked inflate, but not actually in combination with gzip. In general I suspect gzip headers are under-tested at the moment.

PR #204, opened 2024-09-28
Reenable deflate::test::insufficient_compress_space under Miri

inahga:
This test passes under Miri. This includes under s390x, [as the original PR that introduced it was making fixes for](https://github.com/memorysafety/zlib-rs/pull/138). So I see no reason to disable it. Maybe it was a left over disable from development?


folkertdev (approved):
No reason to keep it disabled then, thanks!

PR #211, opened 2024-10-03
fix for reading/writing uninitialized memory

folkertdev:
The issues were in unreleased code, so there is not much of a risk here. Glad we caught it though.

I've also fixed the rest of the `test-libz-rs-sys` tests so that they can run under miri. I'll add some CI infra for that in a followup (probably requires some fighting with github CI).

cc @inahga


inahga (approved):
Thanks!

PR #217, opened 2024-10-07
set `head.done = -1` if a gzip header was configured but the input is not gzip data

folkertdev:


zlib-rs/src/inflate.rs
```diff
@@ -575,9 +575,13 @@ impl<'a> State<'a> {
...
 
+        if let Some(header) = &mut self.head {
+            header.done = -1;
+        }
+
         // check if zlib header is allowed
-        if (self.wrap & 1) != 0
-            && ((self.bit_reader.bits(8) << 8) + (self.bit_reader.hold() >> 8)) % 31 != 0
+        if (self.wrap & 1) == 0
+            || ((self.bit_reader.bits(8) << 8) + (self.bit_reader.hold() >> 8)) % 31 != 0
```
folkertdev, line 584: the C version is messy

```c
            if (!(state->wrap & 1) ||   /* check if zlib header allowed */
#else
            if (
#endif
                ((BITS(8) << 8) + (hold >> 8)) % 31) {
                SET_BAD("incorrect header check");
                break;
            }
```

but I believe this is correct now

Issue #115, opened 2024-06-16
Consider respecting cfg!(fuzzing)


fintelia:
The [`cfg!(fuzzing)` flag](https://rust-fuzz.github.io/book/cargo-fuzz/guide.html#cfgfuzzing) is designed as a way to disable checksums validation and similar checks when Rust code is run under fuzzing. This would be helpful to enable differential fuzzing between the many existing Rust zlib implementations.

folkertdev:
I don't really follow why calculating checksums would be a problem for fuzzing? the checksum is entirely deterministic, and other implementations should generate the same checksum for the same input bytes. Can you elaborate on what exactly we should do differently when `cfg!(fuzzing)` is enabled?

fintelia:
This would be for the inflate codepath. There the fuzzer would be generating random sequences of bytes and feeding them to the library as zlib bitstreams. Most (at least initially) would be corrupt for one reason or another, but over time the fuzzer would start to find bitstreams that we valid other than the checksum. Rather than forcing the fuzzer to also look for valid checksums, you can instead have the decompressor return success even when the checksum doesn't match.


[This is where miniz_oxide ignores the checksum](https://github.com/Frommi/miniz_oxide/blob/d1de8dba2e2bbea6452c9a1d78b221a0f41dadd2/miniz_oxide/src/inflate/core.rs#L1798-L1800) and [here](https://github.com/image-rs/fdeflate/blob/main/fuzz/fuzz_targets/inflate.rs) is a fuzz test that found multiple bugs in fdeflate and miniz_oxide by comparing their behavior to each other.

folkertdev:
that's interesting. I think a lot of code paths are already covered by the raw (so, no header, no footer) deflate stream but control flow does branch on the raw/zlib/gzip mode. 

So, I just naively tried this, and it doesn't seem to do anything?

```
> cargo +nightly fuzz run inflate_random_bytes

warning: unexpected `cfg` condition name: `fuzzing`
   --> /home/folkertdev/rust/zlib-rs/zlib-rs/src/inflate.rs:901:22
    |
901 |             if !cfg!(fuzzing) {
    |                      ^^^^^^^
```

so, there must be more to it than that? But the docs suggest to me that the above should just work.

fintelia:
That is unfortunately expected on recent nightlies. The compiler now "helpfully" prints that even though the fuzzer enables the cfg!(fuzzing) flag

folkertdev:
[...]

PR #231, opened 2024-10-18
fuzzing: inflate fuzz suite for catching UB

inahga:
This suite is mainly intended to catch memory safety bugs and undefined behavior on the inflate code path. To that end it eschews the correctness checks present in other fuzz targets.

It also supports easily running Miri against the generated corpus, to catch any Rust-specific UB that wouldn't be caught by the other LLVM sanitizers.

### DIFF
--- a/fuzz/fuzz_targets/uncompress2.rs
+++ b/fuzz/fuzz_targets/uncompress2.rs
@@ -3,13 +3,23 @@
 //! coverage by disabling checksum validation and disregarding correctness of the results.
 //!
 //! This test must be run with `--features disable-checksum`. It's also suggested to initialize
-//! fuzzing with a corpus of real zlib/gzip files.
+//! fuzzing with a corpus of real zlib/gzip files. Place the corpus in directory
+//! `corpus/uncompress2` (the default corpus location).
+//!
+//! Then, the fuzzer can be run like:
+//!
+//! ```
+//! cargo fuzz run uncompress2 --features disable-checksum -j$(nproc)
+//! ```
+//!
+//! If not starting with an initial corpus, consider using the `-- -max_len=1048576` argument to
+//! test larger inputs.
 //!
 //! libfuzzer uses LLVM sanitizers to detect some classes of bugs and UB. For detecting
 //! Rust-specific UB, use Miri. Once a corpus with suitable coverage has been built, you can run
 //! Miri against the corpus by executing:
 //! ```
-//! MIRIFLAGS=-Zmiri-disable-isolation cargo miri nextest run --bin uncompress2
+//! MIRIFLAGS=-Zmiri-disable-isolation cargo miri nextest run --bin uncompress2 --features disable-checksum
 //! ```
 //! This assumes the corpus is located in the default directory of `corpus/uncompress2`. If it
 //! isn't, specify the corpus directory with the `ZLIB_RS_CORPUS_DIR` environment variable.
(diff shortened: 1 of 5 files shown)


zlib-rs/src/inflate.rs
```diff
@@ -914,7 +914,7 @@ impl<'a> State<'a> {
     }
 
     fn check(&mut self) -> ReturnCode {
-        if self.wrap != 0 {
+        if !cfg!(feature = "__internal-fuzz-disable-checksum") && self.wrap != 0 {
```
inahga, line 917: I think this incidentally closes https://github.com/trifectatechfoundation/zlib-rs/issues/115, FYI @fintelia.

fuzz/fuzz_targets/uncompress2.rs
```diff
@@ -0,0 +1,118 @@
...
+            15 + 32, // Support both zlib and gzip files.
+            zlibVersion(),
+            size_of::<z_stream>() as _,
+        )
+    };
+    assert_eq!(ReturnCode::from(err), ReturnCode::Ok);
+
+    let mut extra = vec![0; 64];
+    let mut name = vec![0; 64];
+    let mut comment = vec![0; 64];
```
inahga, line 45: Maybe we can vary these buffer sizes, or whether they're present at all, but can save that for a later improvement.

inahga:
Sorry, this needs more work on CI.

folkertdev:
can you share how you run this locally (and how we might run it with e.g. ossfuzz). I'm assuming you allow generating larger input sizes, and potentially other settings?

inahga:
I fleshed out the documentation a bit more to have the command I've been using.

The exercise of getting a corpus of gzip files is left to the reader. My strategy was to clumsily assemble a corpus by downloading a bunch of random gzips in GitHub public respositories.

Here's what I did roughly with the `gh` CLI.
```bash
# I think this starts failing earlier than 40 iterations, due to github search limits.
for i in $(seq 40); do gh api "search/code?page=${i}&per_page=100&q=extension%3A.gz+size%3A%3C%3D1048576" | jq -r '.items[].url' >>files.txt; sleep 30; done

# Prune out one gzip file per github organization, since people seem to upload duplicate files.
cat files.txt | sort -u -t/ -k5,5 >files2.txt

# Actually download the files, using your auth token so that you don't get rate limited.
for f in $(cat files2.txt); do curl -O -H"Authorization: Bearer $(gh auth token)" "$(curl -H"Authorization: Bearer $(gh auth token)" -H "Accept: application/vnd.github+json" "$f" | jq -r '.download_url')"; sleep 2; done
```

If you start with the initial corpus, `-max_len` isn't necessary, otherwise you'll probably want to do `-max_len=some number of bytes` because it defaults to 4096 bytes.

---

> and how we might run it with e.g. ossfuzz

I haven't looked at integration with ossfuzz too much, [but it looks fairly straightforward](https://google.github.io/oss-fuzz/getting-started/new-project-guide/rust-lang/). Though I think for coverage purposes we might want to rewrite fuzz targets as C programs and submit it as a C project. Either way you basically you hand them a dockerfile that builds the fuzz binaries.

I haven't looked into it deeply since they say they'll only accept new projects if they "[have a significant user base and/or be critical to the global IT infrastructure](https://google.github.io/oss-fuzz/getting-started/accepting-new-projects/)". Which I don't think qualifies zlib-rs at this precise moment, so I didn't see a reason to complicate running the fuzzer just yet.

fuzz/fuzz_targets/uncompress2.rs
```diff
@@ -0,0 +1,128 @@
...
+use libz_rs_sys::{
+    gz_header, inflate, inflateEnd, inflateGetHeader, inflateInit2_, z_stream, zlibVersion,
+};
+use zlib_rs::{InflateFlush, ReturnCode};
+
+fuzz_target!(|input: &[u8]| -> Corpus { run(input) });
+
+fn run(input: &[u8]) -> Corpus {
+    if input.is_empty() {
+        return Corpus::Reject;
```
folkertdev, line 38: TIL that this is a thing. This is very useful!

folkertdev (commented):
excellent work!

PR #236, opened 2024-10-24
crc32: mark some unsafe functions properly, add some safety documentation

inahga:
Some small nits I encountered in reviewing this code. Shoudn't be any actual functional changes.

- Any function/method that invokes an intrinsic on aarch64 or x86_64 is unsafe, because it has the precondition of ensuring the runtime processor supports the instructions, otherwise the result is UB.
  
  Results in a bit of an annoying proliferation of `unsafe` in `pclmuqdg::Accumulator`. There may be some fiddling that can be done to reduce the `unsafe` footprint, but it doesn't seem like a big deal for internal code.

- [I set `warn(unsafe_op_in_unsafe_fn)`](https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html#details), since otherwise unsafe operations can "hide" in unsafe blocks and thus not receive scrutiny.

- Added a couple of SAFETY comments. I eschewed safety comments for things that struck me as obvious, e.g. taking pointer to a slice that's just fed into an intrinsic.

(@folkertdev let me know if this kind of PR helps. This is all just techniques I was doing to review anyway, so figured they would be useful upstream too).

### DIFF
--- a/zlib-rs/src/crc32.rs
+++ b/zlib-rs/src/crc32.rs
@@ -1,3 +1,4 @@
+#![warn(unsafe_op_in_unsafe_fn)]
 use crate::CRC32_INITIAL_VALUE;
 
 #[cfg(target_arch = "aarch64")]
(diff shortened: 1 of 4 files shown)


folkertdev (approved):
this is very helpful, thanks!

PR #241, opened 2024-10-30
deflate: safety documentation, relax safety requirements of wasm intrinsics

inahga:
Mostly documentation and noise from `#![warn(unsafe_op_in_unsafe_fn]`.

For WASM intrinsics, it is not UB if the system doesn't support the target feature, because WASM interpreters simply won't load binaries that contain unknown instructions. So we can relax the safety requirements accordingly.

### DIFF
--- a/zlib-rs/src/deflate.rs
+++ b/zlib-rs/src/deflate.rs
@@ -1,3 +1,4 @@
+#![warn(unsafe_op_in_unsafe_fn)]
 use core::{ffi::CStr, marker::PhantomData, mem::MaybeUninit, ops::ControlFlow};
 
 use crate::{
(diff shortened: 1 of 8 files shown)


zlib-rs/src/deflate.rs
```diff
@@ -2732,6 +2753,7 @@ pub fn compress_slice_with_flush<'a>(
     config: DeflateConfig,
     flush: DeflateFlush,
 ) -> (&'a mut [u8], ReturnCode) {
+    // SAFETY: a [u8] is a valid [MaybeUninit<u8>].
```
folkertdev, line 2756:
```suggestion
    // SAFETY: a [u8] is a valid [MaybeUninit<u8>], and `compress_with_flush` never uninitializes previously initialized memory.
```

turning a `&mut T` into `&mut MaybeUninit<T>` is extra dangerous because you can write `MaybeUninit::uninit()` to it in safe code. We don't do that, but it is risky.

folkertdev (approved):

Issue #229, opened 2024-10-15
OOB panic in `deflate_stored` after `deflateParams()`


inahga:
Fuzzing reveals a OOB panic in deflate_stored() on the main branch, when we change the deflate parameters midway through the stream.

This seems somewhat related to both https://github.com/trifectatechfoundation/zlib-rs/issues/218 and https://github.com/trifectatechfoundation/zlib-rs/issues/219,

```rust
#[test]
    fn inahga_4() {
        let mut source = [
            161, 161, 161, 161, 179, 179, 179, 179, 179, 179, 133, 133, 133, 133, 13, 13, 13, 13,
            13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 95, 95, 95, 255, 179, 179, 179, 179, 179, 179,
            179, 179, 179, 43, 179, 8, 179, 133, 133, 133, 133, 13, 13, 122, 13, 13, 13, 13, 13,
            13, 13, 13, 13, 3, 13, 13, 95, 95, 95, 13, 13, 255, 255, 255, 189, 189, 189, 189, 189,
            189, 189, 0, 189, 189, 189, 189, 189, 189, 189, 189, 189, 173, 189, 189, 189, 189, 189,
            189, 189, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255,
            255, 255, 255, 255, 255, 255, 255, 255, 255, 95, 95, 95, 255, 255, 255, 255, 255, 255,
            255, 255, 255, 95, 95, 95, 95, 255, 255, 189, 189, 189, 189, 189, 173, 189, 189, 189,
            189, 189, 189, 189, 255, 255, 245, 255, 255, 255, 255, 255, 189, 255, 255, 255, 255,
            255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 189, 189, 189, 189, 189, 189,
            189, 189, 189, 189, 189, 189, 189, 189, 189, 189, 255, 255, 255, 255, 189, 13, 13, 13,
            13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13,
            125, 125, 125, 125, 125, 39, 39, 39, 39, 39, 39, 39, 39, 39, 39, 39, 39, 39, 39, 39,
[...]

PR #255, opened 2024-12-02
Avoid an out-of-bounds panic in deflate_stored

bjorn3:
When shifting the window down, it is possible for the end of the window to be after the last initialized byte of the buffer. These uninitialized bytes don't affect the output of the compression, so we can safely skip copying them to avoid the panic. This matches the code a bit higher up which also does a window slide.

Fixes https://github.com/trifectatechfoundation/zlib-rs/issues/229


bjorn3:
Missed that PR. Approved it.

folkertdev (approved):

PR #260, opened 2024-12-09
Fix `inflate_fast_help` loop bound

folkertdev:
When you finally spot the thing... :facepalm: we are now faster than zlib-ng for all chunk sizes.

chart (5)

It's only a marginal improvement for chunk size 2**4, but quite significant for what are probably the most common chunk sizes (very small chunk sizes likely indicate io-bound operation).

chart (6)

---

After fixing the bug of exiting the loop too early (first commit), I then made the bound a bit more accurate, and finally refactored the output type of `len_and_friends`. The final commit has no performance impact.

### DIFF
--- a/zlib-rs/src/inflate/bitreader.rs
+++ b/zlib-rs/src/inflate/bitreader.rs
@@ -78,6 +78,11 @@ impl<'a> BitReader<'a> {
     }
 
     #[inline(always)]
+    pub fn bytes_remaining_including_buffer(&self) -> usize {
+        (self.end as usize - self.ptr as usize) + (self.bits_used as usize >> 3)
+    }
+
+    #[inline(always)]
     pub fn need_bits(&mut self, n: usize) -> Result<(), ReturnCode> {
         while (self.bits_used as usize) < n {
             self.pull_byte()?;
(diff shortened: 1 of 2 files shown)


zlib-rs/src/inflate/bitreader.rs
```diff
@@ -77,6 +77,11 @@ impl<'a> BitReader<'a> {
         self.end as usize - self.ptr as usize
     }
 
+    #[inline(always)]
+    pub fn bytes_remaining_including_buffer(&self) -> usize {
+        (self.end as usize - self.ptr as usize) + (self.bits_used as usize >> 3)
```
bjorn3, line 82: Using `/ 8` instead of `>> 3` may be a bit clearer and will optimize to the same assembly.

bjorn3 (approved):
Nice!

zlib-rs/src/inflate/bitreader.rs
folkertdev: we can make a pass over the repo at some point. I think zlib-ng never updated this, and the original zlib that it forked was written at a time when apparently that optimization was not (effectively) guaranteed.

PR #272, opened 2024-12-23
Simplify deflate window

folkertdev:
this will conflict with https://github.com/trifectatechfoundation/zlib-rs/pull/271, but should be simple to fix.

We really don't need this fine-grained initialization: just zero the whole thing out at the start (via the allocator if possible) and get rid of a bunch of bookkeeping.


zlib-rs/src/allocate.rs
```diff
@@ -160,6 +160,14 @@ impl Allocator<'static> {
     };
 }
 
+fn null_is_none<T>(ptr: *mut T) -> Option<*mut T> {
```
bjorn3, line 163: Would it make sense to return `Option<NonNull<T>>` instead? Then it would just be a `NonNull::new()` to get the `Option` and the return value is half as big.

zlib-rs/src/deflate/window.rs
```diff
@@ -91,52 +91,9 @@ impl<'a> Window<'a> {
...
+    pub fn initialize_out_of_bounds(&mut self) {}
 
-    pub fn initialize_at_least(&mut self, at_least: usize) {
-        let end = at_least.clamp(self.high_water, self.buf.len());
-        self.buf.as_mut_slice()[self.high_water..end].fill(0);
-
-        self.high_water = end;
-        self.filled = end;
-    }
+    pub fn initialize_at_least(&mut self, at_least: usize) {}
```
bjorn3: This commit in isolation would cause the `.filled()` method to return a shorter output as you removed `self.filled = end;`. Because of the next commit the PR as a whole doesn't have this problem though.

bjorn3 (approved):
with out without the `NonNull` change.

PR #275, opened 2025-01-03
Begin converting match offsets in struct State to u16

brian-pane:
This is the first step of a refactoring to use smaller fields where possible in struct State for better cache locality.

This patch uses the `Pos` type from `longest_match.rs` to represent positions in the window.

My benchmark testing shows no change in performance compared to the current `main`.


folkertdev (approved):
For me there appear to be some minor regressions in instruction count, but that's not something I'm worried about.

PR #277, opened 2025-01-04
Rearrange fields in struct State for cache locality

brian-pane:
On my test system, this yields a small improvement in CPU cycles,
```
Benchmark 1 (68 runs): ./blogpost-compress-baseline-native 1 rs silesia-small.tar
  measurement          mean ± σ            min … max           outliers         delta
  wall_time          73.8ms ± 1.80ms    72.0ms … 86.1ms          1 ( 1%)        0%
  peak_rss           26.6MB ± 84.6KB    26.5MB … 26.7MB          0 ( 0%)        0%
  cpu_cycles          279M  ± 1.11M      278M  …  287M           2 ( 3%)        0%
  instructions        568M  ±  234       568M  …  568M           1 ( 1%)        0%
  cache_references    265K  ± 4.64K      262K  …  298K           7 (10%)        0%
  cache_misses        233K  ± 6.59K      207K  …  244K           6 ( 9%)        0%
  branch_misses      2.90M  ± 4.04K     2.89M  … 2.91M           1 ( 1%)        0%
Benchmark 2 (69 runs): ./target/release/examples/blogpost-compress 1 rs silesia-small.tar
  measurement          mean ± σ            min … max           outliers         delta
  wall_time          73.0ms ±  884us    71.4ms … 77.8ms          6 ( 9%)          -  1.0% ±  0.6%
  peak_rss           26.6MB ± 97.7KB    26.5MB … 26.7MB          0 ( 0%)          +  0.0% ±  0.1%
  cpu_cycles          277M  ± 1.76M      276M  …  290M           1 ( 1%)          -  0.9% ±  0.2%
  instructions        568M  ±  269       568M  …  568M           2 ( 3%)          +  0.0% ±  0.0%
  cache_references    265K  ± 5.08K      262K  …  303K           3 ( 4%)          -  0.0% ±  0.6%
  cache_misses        233K  ± 5.87K      210K  …  239K           6 ( 9%)          -  0.0% ±  0.9%
  branch_misses      2.88M  ± 4.12K     2.87M  … 2.89M           0 ( 0%)          -  0.7% ±  0.0%
```

PR #278, opened 2025-01-06
Add checks for State memory layout

brian-pane:


brian-pane:
Compared to the version in #277, this only checks the first cache line. The reason is that the next 43 and 1/8 cache line boundaries lie in the middle of the various large `TreeDesc` fields in the middle of the `State` structure. If I'm able to move those fields to the end without any performance regression in a later PR, I'll add the rest of the cache line markers in between the small fields.

folkertdev (approved):

PR #279, opened 2025-01-07
Move the TreeDesc fields and finish adding cache line markers to State

brian-pane:
No performance regression measured at compression levels 1-9 on my test system

PR #280, opened 2025-01-08
troubleshooting a ci problem

folkertdev:
in #279, ci fails for unclear reasons; we can't reproduce it locally, so here I'll try to reproduce that PR with tiny incremental changes, and see if we can learn more.


bjorn3:
Let me try clearing the CI cache. Maybe it got corrupted.

folkertdev:
found the issue: when using `libz-sys` on version 1.1.20, the hash_calc function behaves differently from the newer 1.1.21, causing the test failures. We could not reproduce locally because our `Cargo.lock` still used the older version.

folkertdev:
fixed by https://github.com/trifectatechfoundation/zlib-rs/pull/282

PR #283, opened 2025-01-09
Remove the padding from the State structure

brian-pane:
This is a cleanup following the recent restructuring of fields.


brian-pane:
On my test system, the performance of this patch is the same as main for compression levels 1-9.

zlib-rs/src/deflate.rs
```diff
@@ -1226,7 +1216,7 @@ impl<'a> BitWriter<'a> {
     }
 }
 
-#[repr(C)]
+#[repr(C, align(64))]
```
folkertdev, line 1219: it's fine to have this, because that is the intention, but the state is always stored on the heap, and we make sure that all allocations are aligned to 64 bytes anyway, so in practice this does not have any effect.

folkertdev (approved):

PR #284, opened 2025-01-10
Optimize away compare256 for short matches

brian-pane:


brian-pane:
I'm posting this for discussion because it produces mixed results on my test system:
- Improvement in cycles and wall time at compression levels 2 and 9
- Regression in instructions (but no change in cycles or wall time) at levels 3-8.

Whether it helps is probably dependent on the CPU and the input data.

zlib-rs/src/deflate/longest_match.rs
```diff
@@ -190,26 +190,35 @@ fn longest_match_help<const SLOW: bool>(
         // - >= 8: scan_end is bounded by `258 - (4 + 2 + 1)`, so an 8-byte read is in-bounds
         // - >= 4: scan_end is bounded by `258 - (2 + 1)`, so a 4-byte read is in-bounds
         // - >= 2: scan_end is bounded by `258 - 1`, so a 2-byte read is in-bounds
+        let mut len = 0;
         unsafe {
-            if best_len < core::mem::size_of::<u32>() {
+            if best_len < core::mem::size_of::<u64>() {
+                let scan_val = u64::from_ne_bytes(
```
brian-pane, line 196: I also tried `if best_len < 32 {` followed by the equivalent AVX2 implementation (similar to how `compare32` processes each 32=byte chunk), but it was slower overall.

folkertdev (approved):
I think a slightly higher instruction count (correctly predicted branches?) is totally acceptable if we get a performance boost, especially because levels 3,4, 7 and 8 are rarely used in practice.

And the gains at level 2 and 9 are very real, so I think this is OK. Thanks!

PR #286, opened 2025-01-15
fuzzer: make sure the buffer is large enough to store the input

folkertdev:
fuzzers are broken on today's nightly, hopefully that will be fixed soon. 

But I ran into this problem on a local run, and it's a simple fix. I'll wait with merging till we can actually run these on CI though.


bjorn3 (approved):
It may be useful to also have a fuzz case for when the output buffer is too small. That this gives exactly the same amount of data as fits in the output buffer and that you can continue getting data if you call into the decompression function again.

folkertdev:
right, we test this a bunch, but not using fuzzers right now

PR #288, opened 2025-01-16
Fix CI

bjorn3:
This updates two very outdated actions.


nmoinvaz:
A dependabot config can keep your actions up to date.

PR #292, opened 2025-01-30
manual struct target feature

folkertdev:
this essentially inlines all avx2 codepaths.

this basically gives the same performance as `-Ctarget-cpu=native` on my machine.

but it's not the prettiest thing. I'm open to suggestions.

### DIFF
--- a/zlib-rs/src/inflate/writer.rs
+++ b/zlib-rs/src/inflate/writer.rs
@@ -2,6 +2,7 @@ use core::fmt;
 use core::mem::MaybeUninit;
 use core::ops::Range;
 
+use crate::cpu_features::CpuFeatures;
 use crate::weak_slice::WeakSliceMut;
 
 pub struct Writer<'a> {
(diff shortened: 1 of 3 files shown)


folkertdev:
I tried specializing for avx512, but the `target_feature(enable = "avx512f")` etc are not stable, so the avx512 code would just use SSE loads, and that turns out to be worse. So, avx2 is the best we can do right now (until we start looking at avx512 seriously).

zlib-rs/src/inflate/writer.rs
```diff
@@ -140,6 +163,25 @@ impl<'a> Writer<'a> {
...
+                self.copy_match_help::<core::arch::x86_64::__m256i>(offset_from_end, length)
+            }
+            _ => self.copy_match_fallback(offset_from_end, length),
+        }
+    }
+
+    pub fn copy_match_fallback(&mut self, offset_from_end: usize, length: usize) {
         // NOTE: the dynamic check for avx512 makes avx2 slower. Measure this carefully before re-enabling
         //
         //        #[cfg(target_arch = "x86_64")]
```
bjorn3, line 187: The `is_enabled_avx2` branch can be removed from this function, right? If avx2 was supported, `copy_match_dispatch` would not call `copy_match_fallback`.

bjorn3 (approved):
Makes sense. Maybe remove the last two commits though?

zlib-rs/src/inflate/writer.rs
folkertdev: not quite, we still use the `NONE` version in the logic of `len_and_friends`

PR #294, opened 2025-02-17
apply the `crc` target feature to the acle crc32 implementation

folkertdev:
The target feature is required for inlining the intrinsics on `aarch64-unknown-linux-gnu`. For the apple targets it looks like the crc feature is on by default and therefore this change won't have any effect there.

I also removed support for `arm`, which was not hooked up (and I believe the related target features are unstable).


zlib-rs/src/crc32/acle.rs
```diff
@@ -24,28 +25,7 @@ pub unsafe fn crc32_acle_aarch64(crc: u32, buf: &[u8]) -> u32 {
...
-    for w in middle {
-        c = unsafe { __crc32w(c, *w) };
-    }
-
-    c = remainder(c, after);
-
-    !c
-}
-
+#[inline]
```
bjorn3, line 28:
```suggestion
#[inline]
#[target_feature(enable = "crc")]
```

bjorn3 (approved):

PR #299, opened 2025-02-21
Clippy 1.85.0 fixes

folkertdev:


bjorn3:
Could you do the reformatting in a separate commit?

zlib-rs/src/allocate.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unpredictable_function_pointer_comparisons)]
```
folkertdev, line 1: @bjorn3 is this actually risky? We try to be careful in this file about only exposing the function addresses in one way, and miri is OK with the current implementation, but maybe there is still a risk of this misbehaving on some platform?

bjorn3: If the comparison incorrectly fails at worst it would be slower. If the comparison succeeds incorrectly, that means the target function either has UB or is functionally equivalent at the assembly level to whichever function we tried to compare against, so it should be fine too.

folkertdev:
> Could you do the reformatting in a separate commit?

apparently not? it looks like just `cargo fmt` also adds the additional parens that clippy asked for.

bjorn3 (approved):

PR #304, opened 2025-02-25
Store mode on the stack in dispatch

bjorn3:
This improves performance for chunk size 2^4 by a couple percent.

### DIFF
--- a/zlib-rs/src/inflate.rs
+++ b/zlib-rs/src/inflate.rs
@@ -844,6 +844,7 @@ impl State<'_> {
     }
 
     fn dispatch(&mut self) -> ReturnCode {
+        // Note: All early returns must save mode into self.mode again.
         let mut mode = self.mode;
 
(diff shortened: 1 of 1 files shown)


bjorn3:
image

zlib-rs/src/inflate.rs
```diff
@@ -862,12 +844,37 @@ impl State<'_> {
     }
 
     fn dispatch(&mut self) -> ReturnCode {
```
folkertdev, line 846: just to be sure, let's add a comment around the top somewhere warning that early returns must restore the `mode`

folkertdev (approved):
with the comment, and maybe also after running the CI benchmarks to see if anything weird happens there?

zlib-rs/src/inflate.rs
bjorn3: done

Issue #309, opened 2025-02-27
runtime error when no allocator is configured


orangecms:
I experimented a bit with the zlib-rs crate because I am searching for a decent (de)compression library for use in firmware, specifically https://github.com/oreboot/oreboot.

In the end, I ran into https://github.com/trifectatechfoundation/zlib-rs/blob/aee15031daa7178f1a65f8460b1cbb9e3a670742/zlib-rs/src/inflate.rs#L2151-L2153 IIUC - i.e., a runtime error.

This could be checked at compile time instead, I think; is the current way it happens intentional?

Should we add something to the README to tell that an allocator is needed (I'd file a PR)?

Is there a possible intention to offer integration without heap allocation at some point?

folkertdev:
This error is expected, but also it is already possible to provide a custom allocator. Here is some old code that runs with `no_std` on embedded hardware: https://github.com/folkertdev/no-std-zlib-rs. zlib-rs has changed so it may not be fully usable with the latest version, but it should be quite close.

this section

https://github.com/folkertdev/no-std-zlib-rs/blob/05c97c32a0639765bc6f1b3bdafdb45c6ec2ae62/src/main.rs#L79-L81

sets the allocator.

Hopefully that gets you unstuck? 

We do check on CI that we can compile with `no_std`, but otherwise we've not been considering embedded use cases that much, because zlib-rs is quite large in terms of code size. If this works out for you tough, we'd be happy to add some additional documentation on use in `no_std` contexts.

orangecms:
Oh thanks for the quick reply! For more details, I just did some more research and left my notes in https://github.com/oreboot/oreboot/issues/772.

My goal is to avoid an allocator if possible, and we only need decompression on the target machine, so I would expect code size not to become too much of an issue; I just did a check and my code grows from 34K to 87K.
That is absolutely fine, because our use case is application processors / SoCs where we would run Linux, and we need to (de)compress the kernel to fit in a SPI flash, so a few dozen kilobytes of code vs megabytes of kernel size are a good tradeoff.

Anyway, my question is really whether it is a design decision to only support environments that have an allocator, which is all fine; I would just suggest to error at build time then, because I found the runtime error rather confusing, and it could be avoided, I think.

[...]

PR #317, opened 2025-03-14
explicitly document memory management

folkertdev:
fixes https://github.com/trifectatechfoundation/zlib-rs/issues/309

### DIFF
--- a/libz-rs-sys/README.md
+++ b/libz-rs-sys/README.md
@@ -88,6 +88,32 @@ let inflated = &output[..strm.total_out as usize];
 assert_eq!(inflated, input.as_bytes())
 ```
 
+## Memory Management
+
+Zlib does not assume a global allocator, but instead accepts `zalloc` and `zfree` functions for allocating and deallocating memory.
+
+```rust,ignore
+type alloc_func = unsafe extern "C" fn(*mut c_void, c_uint, c_uint) -> *mut c_void;
+type free_func = unsafe extern "C" fn(*mut c_void, *mut c_void);
+
+struct z_stream {
+    // ...
+    zalloc: Option<alloc_func>,
+    zfree: Option<free_func>,
+    opaque: *mut c_void,
+    // ..
+}
+```
+
+These functions can run arbitrary logic, so long as they satisfy their contract. This mechanism can be used e.g. on embedded systems to use an array in static memory as the working memory for (de)compression. For more information see the [`z_stream`](https://docs.rs/zlib-rs/latest/zlib_rs/c_api/struct.z_stream.html) documentation.
+
+For convenience we provide two default memory allocators:
+
+- `rust` uses the rust global allocator
+- `c` uses `malloc` and `free`
+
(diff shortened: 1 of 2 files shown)


libz-rs-sys/README.md
```diff
@@ -88,6 +88,31 @@ let inflated = &output[..strm.total_out as usize];
...
+Zlib does not assume a global allocator, but instead accepts `zalloc` and `zfree` functions for allocating and deallocating memory.
+
+```rust,ignore
+pub type alloc_func = unsafe extern "C" fn(*mut c_void, c_uint, c_uint) -> *mut c_void;
+pub type free_func = unsafe extern "C" fn(*mut c_void, *mut c_void);
+
+struct z_stream {
+    // ...
+    pub zalloc: Option<alloc_func>,
+    pub zfree: Option<free_func>,
```
bjorn3, line 102:
```suggestion
type alloc_func = unsafe extern "C" fn(*mut c_void, c_uint, c_uint) -> *mut c_void;
type free_func = unsafe extern "C" fn(*mut c_void, *mut c_void);

struct z_stream {
    // ...
    zalloc: Option<alloc_func>,
    zfree: Option<free_func>,
```

The pub are not all that useful IMO and kind of implied by the fact that this is documentation. It was inconsistent with the `z_stream` type too.

bjorn3: Also maybe document the `opaque` field too.

zlib-rs/src/c_api.rs
```diff
@@ -18,6 +18,39 @@ pub type voidp = *mut c_void;
...
+///
+/// # Custom allocators
+///
+/// The low-level API supports passing in a custom allocator as part of the [`z_stream`]:
+///
+/// ```no_check
+/// struct z_stream {
+///     // ...
+///     pub zalloc: Option<unsafe extern "C" fn(_: *mut c_void, _: c_uint, _: c_uint) -> *mut c_void>,
+///     pub zfree: Option<unsafe extern "C" fn(_: *mut c_void, _: *mut c_void)>,
```
bjorn3, line 31:
```suggestion
///     pub zalloc: Option<unsafe extern "C" fn(*mut c_void, c_uint, c_uint) -> *mut c_void>,
///     pub zfree: Option<unsafe extern "C" fn(*mut c_void, *mut c_void)>,
```

Rustc has been rejecting argument names in function pointer types since forever.

libz-rs-sys/README.md
```diff
@@ -88,6 +88,31 @@ let inflated = &output[..strm.total_out as usize];
...
+```
+
+These functions can run arbitrary logic, so long as they satisfy their contract. This mechanism can be used e.g. on embedded systems to use an array in static memory as the working memory for (de)compression.
+
+For convenience we provide two default memory allocators:
+
+- `rust` uses the rust global allocator
+- `c` uses `malloc` and `free`
+
+When used as a rust crate, the default is to use the rust global allocator. This behavior can be overridden using [feature flags](#features) to either configure the C allocator as the default, or to configure no default allocator at all. If no default is configured, and no custom allocation and deallocation functions are provided, initialization will return an error.
```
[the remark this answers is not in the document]
folkertdev: right. Because this is (also) the readme, we can only link to the docs.rs version I believe. That'll have to do.

bjorn3 (approved):

PR #324, opened 2025-03-28
add `zlibCompileFlags`

folkertdev:


libz-rs-sys/src/lib.rs
```diff
@@ -91,8 +91,12 @@ const _: () =
 // In spirit this type is `libc::off_t`, but it would be our only libc dependency, and so we
 // hardcode the type here. This should be correct on most operating systems. If we ever run into
 // issues with it, we can either special-case or add a feature flag to force a particular width
+#[cfg(not(target_arch = "wasm32"))]
 pub type z_off_t = c_long;
 
+#[cfg(target_arch = "wasm32")]
+pub type z_off_t = i64;
+
```
folkertdev, line 99: while we're making breaking changes, we may as well fix this too

libz-rs-sys/src/lib.rs
```diff
@@ -1673,6 +1673,113 @@ pub const extern "C" fn zlibVersion() -> *const c_char {
...
+    set_bit!(9, false); // ASMV || ASMINF
+    set_bit!(10, false); // ZLIB_WINAPI
+
+    // One-time table building:
+    set_bit!(12, false); // BUILDFIXED
+    set_bit!(13, false); // DYNAMIC_CRC_TABLE
+
+    // Library content (indicates missing functionality):
+    set_bit!(16, false); // NO_GZCOMPRESS
+    set_bit!(17, false); // NO_GZIP
```
bjorn3, line 1769: This should be true until https://github.com/trifectatechfoundation/zlib-rs/pull/326 gets merged, right?

folkertdev: no this is about gzip support for (de)compression, not the gz file api of #326

bjorn3 (approved):

PR #326, opened 2025-03-29
Partial implementation of gzopen and gzclose

brian-pane:
This includes skeletal implementations of the functions, plus build support (wrapped in a new feature flag called "gz") and test scaffolding.

### DIFF
--- a/libz-rs-sys/Cargo.toml
+++ b/libz-rs-sys/Cargo.toml
@@ -19,6 +19,8 @@ export-symbols = [] # whether the zlib api symbols are publicly exported
 custom-prefix = ["export-symbols"] # use the LIBZ_RS_SYS_PREFIX to prefix all exported symbols
 testing-prefix = ["export-symbols"] # prefix all symbols with LIBZ_RS_SYS_TEST_ for testing
 semver-prefix = ["export-symbols"] # prefix all symbols in a semver-compatible way
+gz = ["dep:libc"] # gzip support has to be configued explicitly, to avoid a libc dependency in the core zlib-rs
 
 [dependencies]
 zlib-rs = { workspace = true, default-features = false }
+libc = { version = "0.2.171", optional = true }
--- a/libz-rs-sys/Cargo.toml
+++ b/libz-rs-sys/Cargo.toml
@@ -19,7 +19,7 @@ export-symbols = [] # whether the zlib api symbols are publicly exported
 custom-prefix = ["export-symbols"] # use the LIBZ_RS_SYS_PREFIX to prefix all exported symbols
 testing-prefix = ["export-symbols"] # prefix all symbols with LIBZ_RS_SYS_TEST_ for testing
 semver-prefix = ["export-symbols"] # prefix all symbols in a semver-compatible way
-gz = ["dep:libc"] # gzip support has to be configued explicitly, to avoid a libc dependency in the core zlib-rs
+gz = ["dep:libc"] # gzip support has to be configured explicitly, to avoid a libc dependency in the core zlib-rs
 
 [dependencies]
 zlib-rs = { workspace = true, default-features = false }
--- a/libz-rs-sys/src/gz.rs
+++ b/libz-rs-sys/src/gz.rs
@@ -4,7 +4,7 @@ pub use zlib_rs::c_api::*;
 use core::ffi::{c_char, c_int, c_uint, CStr };
 use core::mem::size_of;
 use core::ptr;
-use libc::{O_APPEND, O_CLOEXEC, O_CREAT, O_EXCL, O_RDONLY, O_TRUNC, O_WRONLY, SEEK_CUR, SEEK_END};
+use libc::{O_APPEND, O_CREAT, O_EXCL, O_RDONLY, O_TRUNC, O_WRONLY, SEEK_CUR, SEEK_END};
 use zlib_rs::deflate::Strategy;
 
 /// For compatibility with the zlib C API, this structure exposes just enough of the
@@ -113,19 +113,18 @@ unsafe fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gz
     state.strategy = Strategy::Default;
     state.direct = false;
 
-    let mut cloexec = false;
     let mut exclusive = false;
     let mode = CStr::from_ptr(mode);
     for &ch in mode.to_bytes() {
         if ch.is_ascii_digit() {
             state.level = (ch - b'0') as i8;
         } else {
+            // FIXME implement the 'e' flag on platforms where O_CLOEXEC is supported
             match ch {
                 b'r' => state.mode = GzMode::GZ_READ,
                 b'w' => state.mode = GzMode::GZ_WRITE,
                 b'a' => state.mode = GzMode::GZ_APPEND,
                 b'b' => {}, // binary mode is the default
-                b'e' => cloexec = true,
                 b'x' => exclusive = true,
                 b'f' => state.strategy = Strategy::Filtered,
                 b'h' => state.strategy = Strategy::HuffmanOnly,
@@ -162,9 +161,6 @@ unsafe fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gz
         state.fd = fd;
     } else {
         let mut oflag: c_int = 0;
-        if cloexec {
-            oflag |= O_CLOEXEC;
-        }
         if state.mode == GzMode::GZ_READ {
             oflag |= O_RDONLY;
         } else {
--- a/libz-rs-sys/src/gz.rs
+++ b/libz-rs-sys/src/gz.rs
@@ -23,7 +23,7 @@ pub type gzFile = *mut gzFile_s;
 // The internals of a gzip file handle (the thing gzFile actually points to, with the
 // public gzFile_s part at the front for ABI compatibility).
 #[repr(C)]
-struct gz_state {
+struct GzState {
     x: gzFile_s,         // "x" for the exposed part of the data structure
 
     // Fields used for both reading and writing
@@ -100,10 +100,10 @@ unsafe fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gz
         return ptr::null_mut();
     }
 
-    let Some(state) = ALLOCATOR.allocate_zeroed(size_of::<gz_state>()) else {
+    let Some(state) = ALLOCATOR.allocate_zeroed(size_of::<GzState>()) else {
         return ptr::null_mut();
     };
-    let state = state.cast::<gz_state>().as_mut();
+    let state = state.cast::<GzState>().as_mut();
     state.size = 0;
     state.want = GZBUFSIZE;
     state.msg = ptr::null();
@@ -203,14 +203,14 @@ unsafe fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gz
     // FIXME verify the file headers, and initialize the inflate/deflate state
 
     // FIXME change this to core::ptr::from_mut(state).cast::<gzFile_s>() once MSRV >= 1.76
-    (state as *mut gz_state).cast::<gzFile_s>()
+    (state as *mut GzState).cast::<gzFile_s>()
 }
 
-// Deallocate a gz_state structure and all heap-allocated fields inside it.
+// Deallocate a GzState structure and all heap-allocated fields inside it.
 //
 // # Safety
 // The caller must not use the state after passing it to this function.
-unsafe fn free_state(state: &mut gz_state) {
+unsafe fn free_state(state: &mut GzState) {
     if !state.path.is_null() {
         ALLOCATOR.deallocate(state.path.cast_mut(), libc::strlen(state.path) + 1);
     }
@@ -223,7 +223,7 @@ unsafe fn free_state(state: &mut gz_state) {
 /// This function may be called at most once for any file handle.
 #[cfg_attr(feature = "export-symbols", export_name = crate::prefix!(gzclose))]
 pub unsafe extern "C-unwind" fn gzclose(file: gzFile) -> c_int{
-    let Some(state) = file.cast::<gz_state>().as_mut() else {
+    let Some(state) = file.cast::<GzState>().as_mut() else {
         return Z_STREAM_ERROR;
     };
 
--- a/libz-rs-sys/src/gz.rs
+++ b/libz-rs-sys/src/gz.rs
@@ -159,7 +159,7 @@ unsafe fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gz
     // Save the path name for error messages
     // FIXME: support Windows wide characters for compatibility with zlib-ng
     let len = libc::strlen(path) + 1;
-    let Some(path_copy) = ALLOCATOR.allocate_slice_raw::<u8>(len) else {
+    let Some(path_copy) = ALLOCATOR.allocate_slice_raw::<c_char>(len) else {
         free_state(state);
         return ptr::null_mut();
     };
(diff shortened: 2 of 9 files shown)


brian-pane:
With this PR, `gzopen` only opens the file and doesn't yet do any of the other required processing, such as verifying the magic number and reading the file header. I tried to keep the PR simple, to focus on the organization of the code and to make it easier to review. If it looks good, I'll iterate on the remaining functionality in future PRs.

libz-rs-sys/src/gz.rs
```diff
@@ -0,0 +1,230 @@
...
+    // Error information
+    err: c_int,          // last error (0 if no error)
+    msg: *const c_char,  // error message from last error (NULL if none)
+
+    // FIXME: add the zstream field when read/write support is implemented
+}
+
+// Gzip operating modes
+// NOTE: These values match what zlib-ng uses.
+#[derive(PartialEq)]
```
folkertdev, line 62: it's customary to also derive `Eq` when that is sound (basically anything that does not contain floats), so let's do that

libz-rs-sys/src/gz.rs
```diff
@@ -0,0 +1,230 @@
...
+enum GzMode {
+    GZ_NONE = 0,
+    GZ_READ = 7247,
+    GZ_WRITE = 31153,
+    GZ_APPEND = 1,
+}
+
+const GZBUFSIZE: usize = 128 * 1024;
+
+const Z_DEFAULT_COMPRESSION: i8 = -1;
```
folkertdev, line 72: you can use the constant from the crate itself instead (so `crate::Z_DEFAULT_COMPRESSION as i8`)

libz-rs-sys/src/gz.rs
```diff
@@ -0,0 +1,230 @@
+pub use zlib_rs::c_api::*;
+
+use core::ffi::{c_char, c_int, c_uint, c_void, CStr };
+
+use std::ptr;
```
folkertdev, line 5: prefer `core` over `std` if at all possible.

libz-rs-sys/src/gz.rs
```diff
@@ -0,0 +1,230 @@
...
+use libc::{O_APPEND, O_CLOEXEC, O_CREAT, O_EXCL, O_LARGEFILE, O_RDONLY, O_TRUNC, O_WRONLY, SEEK_CUR, SEEK_END};
+use zlib_rs::deflate::Strategy;
+
+// For compatibility with the zlib C API, this structure exposes just enough of the
+// internal state of an open gzFile to support the gzgetc() C macro.
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct gzFile_s {
```
folkertdev, line 13: Public items also need a doc comment (starting each line with `///`), that shows up in the documentation. It's fine to have some standard comments below that for internal details that are for the reader of the code, but not the user of the type. run `cargo doc --open` to see what the docs look like locally.

libz-rs-sys/src/gz.rs
```diff
@@ -0,0 +1,230 @@
...
+
+const GZBUFSIZE: usize = 128 * 1024;
+
+const Z_DEFAULT_COMPRESSION: i8 = -1;
+
+/// Open a gzip file for reading or writing.
+///
+/// # Safety
+/// The caller must ensure that path and mode point to valid C strings. If the
+/// return value is non-NULL, caller must delete it using only gzclose.
```
folkertdev, line 78:
```suggestion
/// return value is non-NULL, caller must delete it using only [`gzclose`].
```

this makes the function name clickable in the documentation.

libz-rs-sys/src/gz.rs
```diff
@@ -0,0 +1,230 @@
...
+// Allocate a block of memory
+// # Safety: if the returned value is non-NULL, caller must delete it only with gz_free.
+unsafe fn gz_alloc<T>() -> *mut T {
+    libc::calloc(1, size_of::<T>()).cast::<T>()
+}
+
+// Free a block of memory that was allocated with gz_alloc
+// # Safety: caller must ensure that ptr was obtained from a prior call to gz_alloc and that
+//           ptr has not already been passed to gz_free.
+unsafe fn gz_free<T>(ptr: *const T) {
```
folkertdev, line 228: should this not be `*mut T`? that's what `gz_alloc` returns.

libz-rs-sys/src/gz.rs
```diff
@@ -0,0 +1,230 @@
...
+/// # Safety
+/// This function may be called at most once for any file handle.
+#[cfg_attr(feature = "export-symbols", export_name = crate::prefix!(gzclose))]
+pub unsafe extern "C-unwind" fn gzclose(file: gzFile) -> c_int{
+    if file.is_null() {
+        return Z_STREAM_ERROR;
+    }
+
+    // FIXME: once read/write support is added, clean up internal buffers
+    let state = &*file.cast::<gz_state>();
```
folkertdev, line 208: this is a bit safer

```suggestion
    let Some(state) = file.cast::<gz_state>().as_ref() else { 
        return Z_STREAM_ERROR;
    };

    // FIXME: once read/write support is added, clean up internal buffers
```

libz-rs-sys/src/gz.rs
```diff
@@ -0,0 +1,230 @@
...
+        Z_OK
+    } else {
+        Z_ERRNO
+    }
+}
+
+// Allocate a block of memory
+// # Safety: if the returned value is non-NULL, caller must delete it only with gz_free.
+unsafe fn gz_alloc<T>() -> *mut T {
+    libc::calloc(1, size_of::<T>()).cast::<T>()
```
folkertdev, line 222: The caller must also not ask for any `T` whose alignment is larger than what malloc provides (I believe it always aligns to at least 8 (16 on most platforms), so the code here is fine in practice, but it is part of the safety contract.

I'd like to statically assert that fact, but with our minimum supported rust version we can't

libz-rs-sys/src/gz.rs
```diff
@@ -0,0 +1,230 @@
...
+unsafe fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gzFile {
+    if path.is_null() || mode.is_null() {
+        return ptr::null_mut();
+    }
+
+    let state_ptr = gz_alloc::<gz_state>();
+    if state_ptr.is_null() {
+        return ptr::null_mut();
+    }
+    let state = &mut *state_ptr;
```
folkertdev, line 97: like below, this should use `as_mut` in combination with a let-else

folkertdev (commented):
overall this is solid, I've left a bunch of inline comments.

Currently the new files are not tested on CI (for clippy, formatting, and actually running the tests). That means the `gz` flag has to be enabled for instance here https://github.com/trifectatechfoundation/zlib-rs/blob/main/.github/workflows/checks.yaml#L156

brian-pane:
I think the wasm32 run is failing when the gz `open_close` test case tries to open a file under `$CARGO_MANIFEST_DIR/src/test-data`. Is that path available in the wasm32 test environment? I was able to run the test locally using `wasmtime --dir {absolute path to the CARGO_MANIFEST_DIR}` but I don't know how the real CI environment works.

libz-rs-sys/src/gz.rs
```diff
@@ -0,0 +1,239 @@
...
+
+/// For compatibility with the zlib C API, this structure exposes just enough of the
+/// internal state of an open gzFile to support the gzgetc() C macro.
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct gzFile_s {
+    have: c_uint,        // number of bytes available at next
+    next: *const Bytef,  // next byte of uncompressed data
+    pos: u64,            // current offset in uncompressed data stream
+}
```
folkertdev, line 18: I think we should instead make this type opaque to rust users, so similarly to https://docs.rs/libz-sys/1.1.22/libz_sys/enum.gzFile_s.html use an enum with no variants. The `GzFile` can just store these fields, with a comment that the fields must not move

libz-rs-sys/src/gz.rs
```diff
@@ -0,0 +1,239 @@
...
+    // Error information
+    err: c_int,          // last error (0 if no error)
+    msg: *const c_char,  // error message from last error (NULL if none)
+
+    // FIXME: add the zstream field when read/write support is implemented
+}
+
+// Gzip operating modes
+// NOTE: These values match what zlib-ng uses.
+#[derive(Eq, PartialEq)]
```
folkertdev, line 63: nit but generall `PartialEq` comes first

libz-rs-sys/src/gz.rs
folkertdev, line 1: this file (maybe others too) need to be formatted

libz-rs-sys/src/gz.rs
```diff
@@ -0,0 +1,239 @@
...
+/// internal state of an open gzFile to support the gzgetc() C macro.
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct gzFile_s {
+    have: c_uint,        // number of bytes available at next
+    next: *const Bytef,  // next byte of uncompressed data
+    pos: u64,            // current offset in uncompressed data stream
+}
+
+// File handle for an open gzip file.
```
folkertdev, line 20: this should be a doc comment `///`

libz-rs-sys/src/gz.rs
```diff
@@ -0,0 +1,239 @@
...
+/// The caller must ensure that path and mode are NULL or point to valid C strings.
+unsafe fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gzFile {
+    if path.is_null() || mode.is_null() {
+        return ptr::null_mut();
+    }
+
+    let Some(state) = ALLOCATOR.allocate_zeroed(size_of::<GzState>()) else {
+        return ptr::null_mut();
+    };
+    let state = state.cast::<GzState>().as_mut();
```
folkertdev, line 106: I think this is technically unsound because alignment is ignored. (it works out though because malloc aligns to at least 8 in practice). I'll refactor the zeroed allocator apis a bit to make this easier.

folkertdev (commented):
I fixed the wasm CI issue by explicitly passing the path. I've also opened https://github.com/bytecodealliance/wasmtime/issues/10490 because the behavior of `--dir` with non-absolute paths is confusing. 

I've also made sure that the `prefix` macro is not publicly exported (how to do that is not clear at all, but something I've come across before).

libz-rs-sys/src/gz.rs
```diff
@@ -0,0 +1,239 @@
...
+    state.strategy = Strategy::Default;
+    state.direct = false;
+
+    let mut exclusive = false;
+    let mode = CStr::from_ptr(mode);
+    for &ch in mode.to_bytes() {
+        if ch.is_ascii_digit() {
+            state.level = (ch - b'0') as i8;
+        } else {
+            // FIXME implement the 'e' flag on platforms where O_CLOEXEC is supported
```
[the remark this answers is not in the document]
brian-pane: Yes, it looks like Rust libc doesn't provide the constant on platforms where C libc doesn't support it. Here is a simple example:

```
use libc::{O_CREAT, O_LARGEFILE, O_EXCL};

fn main() {
    let _ = O_CREAT | O_LARGEFILE | O_EXCL;
}
```

On modern Linux, this compiles cleanly. On macOS, compilation fails with:
```
error[E0432]: unresolved import `libc::O_LARGEFILE`
 --> src/main.rs:1:21
  |
1 | use libc::{O_CREAT, O_LARGEFILE, O_EXCL};
  |                     ^^^^^^^^^^^ no `O_LARGEFILE` in the root

For more information about this error, try `rustc --explain E0432`.
```

On an older Linux, it will fail with the same error, except for `O_EXCL`.

With C libc, it's easy to conditionally compile:
```
int flags = O_CREAT
#ifdef O_LARGEFILE
    | O_LARGEFILE
#endif
#ifdef O_EXCL
    | O_EXCL
#endif
;
```
I'm not sure if there's an easy way to do that with Rust libc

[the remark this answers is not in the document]
brian-pane: I opened issue [4358](https://github.com/rust-lang/libc/issues/4358) for the libc crate to ask about the missing constants.

libz-rs-sys-cdylib/Cargo.lock
bjorn3, line 1: @folkertdev why is libz-rs-sys-cdylib not a part of the main workspace?

libz-rs-sys/Cargo.toml
```diff
@@ -19,6 +19,8 @@ export-symbols = [] # whether the zlib api symbols are publicly exported
 custom-prefix = ["export-symbols"] # use the LIBZ_RS_SYS_PREFIX to prefix all exported symbols
 testing-prefix = ["export-symbols"] # prefix all symbols with LIBZ_RS_SYS_TEST_ for testing
 semver-prefix = ["export-symbols"] # prefix all symbols in a semver-compatible way
+gz = ["dep:libc"] # gzip support has to be configued explicitly, to avoid a libc dependency in the core zlib-rs
```
bjorn3, line 22: *configured

libz-rs-sys/src/gz.rs
```diff
@@ -0,0 +1,245 @@
...
+
+    // Fields used for seek requests
+    skip: i64,  // amount to skip (already rewound if backwards)
+    seek: bool, // whether a seek request is pending
+
+    // Error information
+    err: c_int, // last error (0 if no error)
+    msg: *const c_char, // error message from last error (NULL if none)
+
+                // FIXME: add the zstream field when read/write support is implemented
```
bjorn3, line 64: Is this weird formatting caused by rustfmt?

libz-rs-sys-cdylib/Cargo.lock
folkertdev: feature flags: the cdylib configures the C allocator by default

libz-rs-sys/src/gz.rs
folkertdev: https://github.com/trifectatechfoundation/zlib-rs/pull/327 should make this `allocator.allocate_zeroed_raw::<GzState>()` and will correctly handle alignment

brian-pane: Thank you! And to allocate space for a C string, I guess it's `allocator.allocate_slice_raw::<u8>(strlen + 1)`?

folkertdev: yes. Then you just need to make sure to write to the memory before you read it, and must write the `NULL` byte. We could add a variant for zeroed slices but I think `slice_allocate_raw` should work fine.

libz-rs-sys/src/gz.rs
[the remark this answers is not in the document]
folkertdev: doesn't the fact that it's listed on this page just mean that linux mandates that the constant exists?

https://man7.org/linux/man-pages/man2/open.2.html

If rust libc does not provide a constant that is available from the system libc, that's a bug in the rust libc, and we can fix it but can't defend ourselves against that.

brian-pane:
I suspect the failure on x86_64-pc-windows-gnu might be because I'm using the Allocator allocate/deallocate interface wrong, but I don't have a way to repro it.

folkertdev:
I suspect the problem is `strncpy` not appending a NULL. You can `printf` the `state.path` and see that it's not right.

folkertdev:
I'll just try debugging on CI a bit. On linux all is well (at least, valgrind reports nothing), so yeah I'm not sure what's up.

folkertdev:
Allright, got it. We were making some assumptions about alignment, which my new code for zeroed allocations violated. I also fixed some other issues with clippy and dropping `CString` values prematurely in the tests.

libz-rs-sys/src/gz.rs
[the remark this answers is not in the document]
brian-pane: The libc maintainers recommended an [autocfg](https://crates.io/crates/autocfg) crate, but that would create an additional external dependency for zlib-rs. For now, I'll try the `[cfg(target_os = "...")]`(in a future PR).

brian-pane:
The three failed CI checks are all due to an external issue:

> This is a scheduled Ubuntu 20.04 brownout. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101

folkertdev (approved):
allright, I think this should be a solid foundation now. I fixed that CI issue, and also added some extra docs to `gzclose`. In general we should, when the functions are complete, basically match the zlib docs (the idea being: users of our library should not have to read the official zlib docs; ours have all of the information too).

PR #328, opened 2025-04-01
fix alignment in zeroed allocations

folkertdev:
Using an incorrect alignment is UB (on some platforms, e.g. miri caught it on windows, but not on linux)

### DIFF
--- a/zlib-rs/src/allocate.rs
+++ b/zlib-rs/src/allocate.rs
@@ -15,6 +15,8 @@ use alloc::alloc::GlobalAlloc;
 #[allow(non_camel_case_types)]
 type size_t = usize;
 
+const ALIGN: u8 = 64;
+
 /// # Safety
 ///
 /// This function is safe, but must have this type signature to be used elsewhere in the library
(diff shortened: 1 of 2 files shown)


zlib-rs/src/allocate.rs
```diff
@@ -251,7 +270,11 @@ impl Allocator<'_> {
     fn allocate_layout_zeroed(&self, layout: Layout) -> *mut c_void {
         #[cfg(feature = "rust-allocator")]
         if self.zalloc == Allocator::RUST.zalloc {
-            return unsafe { std::alloc::System.alloc_zeroed(layout).cast() };
+            let ptr = unsafe { zalloc_rust_calloc(self.opaque, layout.size() as _, 1) };
```
bjorn3, line 273: Why is the input alignment now ignored entirely?

folkertdev: hmm, I'll add some asserts then. There is little we can do about it, but we know that our allocations never have alignments higher than 64.

PR #332, opened 2025-04-02
Add Linux-specific gzopen options and more tests

brian-pane:


brian-pane:
My new test cases try to create temp files and directories, and the wasm32 run is failing with:
```
thread 'main' panicked at library/std/src/sys/pal/wasi/os.rs:267:5:
no filesystem on wasm
```

test-libz-rs-sys/src/gz.rs
```diff
@@ -6,42 +6,83 @@ use std::ffi::CString;
...
-        assert_eq!(gzclose(handle), Z_OK);
-    }
+// Generate a file path relative to a specified directory prefix
+macro_rules! path {
+    ($prefix:expr, $str:expr) => {{
+        let mut path_buf = $prefix.to_path_buf();
+        path_buf.push($str);
+        path_buf.as_path().to_str().unwrap().to_owned()
+    }};
+}
```
folkertdev, line 22: is there a particular reason for the macros? Usually standard functions have better editor support and are easier to reason about.

test-libz-rs-sys/src/gz.rs
```diff
@@ -6,42 +6,83 @@ use std::ffi::CString;
...
+    // Create a temporary directory that will be automatically removed when
+    // temp_dir goes out of scope.
+    let temp_dir_path = if cfg!(target_os = "wasi") {
+        std::path::PathBuf::from("/tmp/")
+    } else {
+        std::env::temp_dir()
+    };
+
+    let temp_dir = tempfile::TempDir::new_in(temp_dir_path).unwrap();
+    let temp_path = temp_dir.path();
```
folkertdev, line 69: so yeah we'll have to use `TempDir::new_in` instead of `TempDir::new` throughout. We can probably extract this logic into a function when we add more tests.

folkertdev (commented):
I got wasm to work for the tests. You need to make `tempfile` not use that `std::env::temp_dir` function

test-libz-rs-sys/src/gz.rs
brian-pane: I made `test_open` a macro so failures in its `assert` calls will be reported by the test driver as happening on the line number where the macro is called. But the other macros can easily be functions. I'll push an update to this PR later today to make them functions.

libz-rs-sys/src/gz.rs
```diff
@@ -120,17 +120,20 @@ unsafe fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gz
...
     for &ch in mode.to_bytes() {
         if ch.is_ascii_digit() {
             state.level = (ch - b'0') as i8;
         } else {
-            // FIXME implement the 'e' flag on platforms where O_CLOEXEC is supported
             match ch {
                 b'r' => state.mode = GzMode::GZ_READ,
                 b'w' => state.mode = GzMode::GZ_WRITE,
                 b'a' => state.mode = GzMode::GZ_APPEND,
                 b'b' => {} // binary mode is the default
```
folkertdev, line 134: this match misses the `+` character from zlib-ng. Is that deliberate?

libz-rs-sys/src/gz.rs
```diff
@@ -172,7 +175,18 @@ unsafe fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gz
...
+        }
+
+        #[cfg(target_os = "linux")]
+        if cloexec {
+            oflag |= libc::O_CLOEXEC;
+        }
+
         if state.mode == GzMode::GZ_READ {
             oflag |= O_RDONLY;
         } else {
```
folkertdev, line 192: I can't comment on it on the line, but I think the block for `if state.mode == GzMode::GZ_APPEND` should be moved out of the else branch, and should also run if the user did provide a file descriptor. At least, that is how I read the C code.

folkertdev: there are some missing comments there as well

folkertdev (approved):
nice, thanks!

PR #333, opened 2025-04-02
Add some comments to gzopen_help and reduce the amount of unsafe code

brian-pane:


libz-rs-sys/src/gz.rs
```diff
@@ -101,15 +101,17 @@ pub unsafe extern "C-unwind" fn gzopen(path: *const c_char, mode: *const c_char)
 ///
 /// # Safety
 /// The caller must ensure that path and mode are NULL or point to valid C strings.
-unsafe fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gzFile {
+fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gzFile {
```
folkertdev, line 104: this function is still unsafe: it has preconditions that the caller must satisfy, that are not verified by the compiler.

libz-rs-sys/src/gz.rs
folkertdev, line 1: let's add this at the top

```rust
#![warn(unsafe_op_in_unsafe_fn)]
```

libz-rs-sys/src/gz.rs
```diff
@@ -133,7 +137,12 @@ unsafe fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gz
                 b'a' => state.mode = GzMode::GZ_APPEND,
                 b'+' => {
                     // Read+Write mode isn't supported
-                    free_state(state);
+                    // Safety: we know state is a valid pointer because it was allocated earlier
+                    // in this function, and it is not used after the free because we return
+                    // immediately afterward.
+                    unsafe {
+                        free_state(state);
+                    }
```
folkertdev, line 145: if you move the `;` to the after of the closing bracket of the unsafe block, this will format on one line, which I personally prefer.

libz-rs-sys/src/gz.rs
```diff
@@ -238,10 +260,12 @@ unsafe fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gz
 //
 // The caller must not use the state after passing it to this function.
 unsafe fn free_state(state: &mut GzState) {
-    if !state.path.is_null() {
-        ALLOCATOR.deallocate::<c_char>(state.path.cast_mut(), libc::strlen(state.path) + 1);
+    unsafe {
```
folkertdev, line 263: here we just promise to only deallocate values that have been allocated with `ALLOCATOR`

libz-rs-sys/src/gz.rs
```diff
@@ -253,17 +277,18 @@ unsafe fn free_state(state: &mut GzState) {
 ///
 /// # Safety
 ///
-/// This function may be called at most once for any file handle.
+/// This function may be called at most once for any file handle. The caller is responsible
+/// for ensuring that the file handle either points to a GzState object or is null.
```
folkertdev, line 281: In the main library we use this phrasing for what "valid" means. It's not just about being able to address the bytes, it also must e.g. by allocated by our allocator.

> /// * Either
> ///     - `file` is `NULL`
> ///     - `file` satisfies the requirements of `&mut *file` and was initialized with [`gzopen`] or similar

libz-rs-sys/src/gz.rs
```diff
@@ -109,7 +111,9 @@ unsafe fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gz
     let Some(state) = ALLOCATOR.allocate_zeroed_raw::<GzState>() else {
         return ptr::null_mut();
     };
-    let state = state.cast::<GzState>().as_mut();
+    // Safety: the allocate_zeroed_raw call above ensures that the allocated block
+    // has the right size and alignment to be used as a GzState.
+    let state = unsafe { state.cast::<GzState>().as_mut() };
```
folkertdev, line 116: A crucial detail here is that it is valid for `GzState` to be all zero bytes. Creating a reference also means we assert that this value is correctly initialized, which the `calloc` here handles for us.

libz-rs-sys/src/gz.rs
brian-pane: I don't understand the part about "`file` satisfies the requirements of `&mut *file`." The requirements of `&mut *file` in this case aren't sufficient, since `file` is a pointer to a `gzFile_s` and we need the caller to pass a pointer to a specific subclass of `gzFile_s`, the `GzState`.

folkertdev: right, maybe it's clearest to say that it must either be `NULL` or the output of `gzopen` (as far as I can tell `gzopen` is the only way to create a `gzFile`)? Given that `gzFile` is opaque, it is just never valid for a user to construct one of these on their own.

brian-pane: Thanks, that makes sense. I'll add a comment saying the argument must have been provided by `gzopen`.

libz-rs-sys/src/gz.rs
```diff
@@ -253,17 +280,18 @@ unsafe fn free_state(state: &mut GzState) {
 ///
 /// # Safety
 ///
-/// This function may be called at most once for any file handle.
+/// - The file handle must have been obtained from a function in this library, such as [`gzopen`].
+/// - This function may be called at most once for any file handle.
```
folkertdev, line 284: from a safety perspective, a `NULL` pointer is also valid

folkertdev: it's not very useful, but it's not UB

folkertdev (approved):
nice, btw for minor formatting fixes and such using a force push (with some `git commit --amend` or similar) is totally fine

PR #334, opened 2025-04-04
Fix unsafe_op_in_unsafe_fn warning in AArch64 CRC32 implementation

brian-pane:


brian-pane:
Did the CI recently upgrade to a newer version of Clippy? There was a lint error on some existing code, but I included a fix for it.

zlib-rs/src/crc32/acle.rs
```diff
@@ -10,7 +10,8 @@ pub unsafe fn crc32_acle_aarch64(crc: u32, buf: &[u8]) -> u32 {
     // SAFETY: [u8; 8] safely transmutes into u64.
     let (before, middle, after) = unsafe { buf.align_to::<u64>() };
 
-    c = remainder(c, before);
+    // SAFETY: `remainder` verifies that `before` contains enough bytes before dereferencing
+    c = unsafe { remainder(c, before) };
```
folkertdev, line 14: that is not what the usafety of `remainder` is about. The function is unsafe only because it uses a target feature, and at least for our minimum supported rust version, that implies that the function must be unsafe. In this caller we also enable/assume the `crc` target feature, and so there are no other preconditions to satisfy.

folkertdev (commented):
yeah a new rust release usually means a couple of new clippy lints, so that happened last week

folkertdev (approved):

PR #335, opened 2025-04-06
Finish implementing gz_open and add gzdopen and gzerror

brian-pane:


brian-pane:
#334 should fix the Clippy error.

brian-pane:
Oops, C string literals of the form `c"example"` aren't supported until Rust 1.77, and the project MSRV is 1.75. I'll replace these with `b"example\0"`.

libz-rs-sys/src/gz.rs
```diff
@@ -90,15 +101,41 @@ compile_error!("Either rust-allocator or c-allocator feature is required");
...
+/// * If successful, an opaque handle that the caller can later free with [`gzfree`]
+/// * On error, a null pointer
+///
+/// # Safety
+///
+/// The caller must ensure that `path` points to a valid C string. If the
+/// return value is non-NULL, caller must delete it using only [`gzclose`].
+#[cfg_attr(feature = "export-symbols", export_name = crate::prefix!(gzdopen))]
+pub unsafe extern "C-unwind" fn gzdopen(fd: c_int, mode: *const c_char) -> gzFile {
+    let path = format!("<fd:{}>\0", fd);
```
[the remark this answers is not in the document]
folkertdev: that simplifies some other logic below too

libz-rs-sys/src/gz.rs
```diff
@@ -249,12 +281,86 @@ unsafe fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gz
...
+unsafe fn gz_error(state: &mut GzState, err: c_int, msg: *const c_char) {
+    if !state.msg.is_null() {
+        // NOTE: zlib-ng skips the deallocation here if state.err == Z_MEM_ERROR.
+        // Instead, we maintain the invariant that `state.msg`, if non-null, is
+        // allocated using `ALLOCATOR`, and we assume that the heap is coherent
+        // enough that we can (and should) free the old value even if a new alloc
+        // just failed.
+        unsafe { deallocate_cstr(state.msg.cast_mut()) };
+        state.msg = ptr::null_mut();
+    }
```
folkertdev, line 327: can you elaborate here? My default is to just do what they do. Why should we deviate?

libz-rs-sys/src/gz.rs
```diff
@@ -249,12 +281,86 @@ unsafe fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gz
...
+    // Safety: `gzopen` and `gzdopen` ensure that `state.path` is a non-null C string.
+    let Ok(path) = (unsafe { CStr::from_ptr(state.path).to_str() }) else {
+        return;
+    };
+    // Safety: the caller of this function is reponsible for ensuring that `msg` is a C string,
+    // and we exit this function above if `msg` is null.
+    let Ok(msg) = (unsafe { CStr::from_ptr(msg).to_str() }) else {
+        return;
+    };
+    let err_msg = format!("{}: {}\0", path, msg);
```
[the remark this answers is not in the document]
folkertdev: So while ugly, I think this should figure out the size, perform the allocation, and use some low-level `ptr::copy` (potentially non-overlapping copy) to compose the msg.

libz-rs-sys/src/gz.rs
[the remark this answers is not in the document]
folkertdev: The implementation in this file should, but tests can use `std`

libz-rs-sys/src/gz.rs
```diff
@@ -249,12 +281,86 @@ unsafe fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gz
...
+    }
+
+    // Format the error string to include the file path.
+    // Safety: `gzopen` and `gzdopen` ensure that `state.path` is a non-null C string.
+    let Ok(path) = (unsafe { CStr::from_ptr(state.path).to_str() }) else {
+        return;
+    };
+    // Safety: the caller of this function is reponsible for ensuring that `msg` is a C string,
+    // and we exit this function above if `msg` is null.
+    let Ok(msg) = (unsafe { CStr::from_ptr(msg).to_str() }) else {
```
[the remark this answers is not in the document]
bjorn3: C strings on Windows don't use UTF-16. They use the same encoding as on Unix.

libz-rs-sys/src/gz.rs
```diff
@@ -249,28 +288,166 @@ unsafe fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gz
...
+
 // Deallocate a GzState structure and all heap-allocated fields inside it.
 //
 // # Safety
 //
 // - The `state` object and all heap-allocated fields within it must have been obtained
 //   using `ALLOCATOR`.
 // - The caller must not use the `state` after passing it to this function.
-unsafe fn free_state(state: &mut GzState) {
+unsafe fn free_state(state: *mut GzState) {
```
folkertdev, line 411: this can't take a mutable reference (MIRI complains if it does. According to rust's model, you can't just destroy a mutable reference).

libz-rs-sys/src/gz.rs
```diff
@@ -249,28 +289,155 @@ unsafe fn gzopen_help(path: *const c_char, fd: c_int, mode: *const c_char) -> gz
...
+    unsafe {
         ALLOCATOR.deallocate::<GzState>(state, 1);
     }
 }
 
+// Free a string that was allocated with `ALLOCATOR`
+//
+// # Safety
+//
+// `s` must be either null or a null-terminated C string that was allocated with `ALLOCATOR`.
```
folkertdev, line 431: the length of the string (plus the null terminator) must also match the length of the allocation

libz-rs-sys/src/gz.rs
```diff
@@ -301,3 +468,255 @@ pub unsafe extern "C-unwind" fn gzclose(file: gzFile) -> c_int {
...
+    let src = unsafe { CStr::from_ptr(src) };
+
+    let len = src.to_bytes_with_nul().len();
+    let Some(dst) = ALLOCATOR.allocate_slice_raw::<c_char>(len) else {
+        return ptr::null_mut();
+    };
+
+    // SAFETY: src and dst don't overlap, because dst was just allocated. src is valid for a read
+    // of len bytes, and dst is valid for a write of len bytes.
+    unsafe { core::ptr::copy_nonoverlapping(src.as_ptr(), dst.as_ptr(), len) };
```
folkertdev, line 549: I'm using the `core::ptr` function here (and not some libc function) so that this runs with Miri

folkertdev (commented):
well this got kind of big. I've made some changes, please review those from your side.

- the state now holds the `Source`, so that `fd_path` is only used when actually formatting an error message
- `fd_path` now uses some `core::fmt` logic, so it's simpler and shorter
- the tests now in `libz-rs-sys` now run with Miri, to make sure we don't introduce any UB with all of these C strings.

libz-rs-sys/src/gz.rs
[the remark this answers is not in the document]
folkertdev: really? that means any syscall involving a `OsString` must perform a conversion right?

In any case, C strings also don't need to be UTF-8, so I think it's best to not convert to `&str`.

libz-rs-sys/src/gz.rs
```diff
@@ -296,72 +296,49 @@ unsafe fn gzopen_help(source: Source, mode: *const c_char) -> gzFile {
...
-//
-// * A pointer to a null-terminated C string that the caller now owns.
-// * Or a null pointer if an error occurs.
-//
-// # Safety
-//
-// The return value, if non-null, must be freed using only `ALLOCATOR`.
-unsafe fn fd_path(fd: c_int) -> *const c_char {
-    // This is equivalent to `format!("<fd:{}>\0", fd)`, but without the dependency on std::format.
+fn fd_path(buf: &mut [u8; 17], fd: c_int) -> &CStr {
```
brian-pane, line 300: The `[u8; 17]` could use a comment to document why it's that size. (6 bytes for `"<fd:>\0"`, 11 bytes for the decimal representation of the smallest possible `i32` value, including the minus sign.)

It's probably also worth documenting that this function will truncate the formatted string (on the left) in the extremely unlikely event that the buffer is too small -- which could happen on an ILP64 system where `c_int` is `i64`, if this code is used in a process that has more than 1e10 open file descriptors.

folkertdev: I've just bumped the buffer size to be sure we never run into issues

PR #336, opened 2025-04-08
bump codecov action version

folkertdev:


.github/workflows/checks.yaml
```diff
@@ -91,7 +91,7 @@ jobs:
           RUST_BACKTRACE: 1
         if: matrix.os != 'windows-2022'
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@eaaf4bedf32dbdc6b720b63067d99c4d77d6047d
+        uses: codecov/codecov-action@v5
```
bjorn3, line 94:
```suggestion
        uses: codecov/codecov-action@0565863a31f2c772f9f0395002a31e3f06189574 # v5.4.0
```

folkertdev: why pin the version? We've also seen a bunch of CI breakage from continuing to use old versions of actions.

bjorn3: To reduce risk of supply chain attacks. For periodic updates we could use something like renovatebot which even supports updating the hash and version in the comment at the same time: https://docs.renovatebot.com/modules/manager/github-actions/#digest-pinning-and-updating

folkertdev: Sure, though the blast radius is really small here (attackers can mess with our code coverage?

bjorn3: In general it would allow making arbitrary changes to the repo contents as the workflow doesn't restrict the permissions of the `GITHUB_TOKEN`. This repo has configured the default permissions to be read-only however, so it is indeed pretty harmless for us.

PR #337, opened 2025-04-09
run the uncompress fuzzer with a custom corpus and gather fuzzer codecov

folkertdev:
The corpus for decompression is generated by https://github.com/folkertdev/compression-corpus/blob/main/.github/workflows/generate.yml, to cover all of the compression levels, plus some extra things. It does pretty well given that it only runs for a limited time.

We could store the corpus zip in the repo, right now it's downloaded by the action.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -254,6 +254,61 @@ jobs:
             RUST_BACKTRACE=1 cargo fuzz run --no-default-features --features="$features" $target -- -max_total_time=10
           done
 
+  fuzz-code-coverage:
+    name: Fuzz with code coverage 
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - fuzz_target: uncompress2
+            corpus: "gzip-files/compressed"
+            features: '--no-default-features --features="disable-checksum"'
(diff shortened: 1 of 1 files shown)


.github/workflows/checks.yaml
```diff
@@ -254,6 +254,62 @@ jobs:
...
+          toolchain: nightly
+          components: llvm-tools-preview
+      - name: Install cargo fuzz & rustfilt
+        uses: taiki-e/install-action@56ab7930c591507f833cbaed864d201386d518a8
+        with:
+          tool: cargo-fuzz,rustfilt
+      - name: Download custom decompression corpus
+        if: ${{ contains(matrix.corpus, 'gzip-files') }}
+        run: |
+          ~/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/bin/llvm-cov show --help
```
bjorn3, line 288: You could use `$(rustc --print sysroot)` instead of hard-coding `~/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu`.

folkertdev:
Yeah probably at this point, I'll set that up

PR #339, opened 2025-04-10
Add gzclearerr and gzeof functions

brian-pane:


libz-rs-sys/src/gz.rs
```diff
@@ -782,4 +846,47 @@ mod tests {
...
+        assert_eq!(unsafe { gzeof(handle) }, 0);
+
+        // gzclearerr should reset the eof and past flags.
+        let state = (unsafe { handle.cast::<GzState>().as_mut() }).unwrap();
+        state.eof = true;
+        assert_eq!(unsafe { gzeof(handle) }, 0);
+        state.past = true;
+        assert_eq!(unsafe { gzeof(handle) }, 1);
+        unsafe { gzclearerr(handle) };
+        assert!(!state.eof);
```
brian-pane, line 871: The Miri test run flags this line as an error because of a "foreign write" to state.eof by gzclearerr. I think that might be because the pointer aliasing and downcasting here has confused Miri, but it could use a second set of eyes.

folkertdev: the aliasing is in violation of rust's rules: `state` is a mutable reference, which implies exclusive access, but its value is read via `handle`. You need to use one or the other, and I suspect here it's easiest to just never create the mutable reference there. I'll post a suggestion

libz-rs-sys/src/gz.rs
```diff
@@ -782,4 +846,47 @@ mod tests {
...
+        assert!(!state.past);
+        assert_eq!(unsafe { gzeof(handle) }, 0);
+
+        // Set an error flag and message.
+        unsafe { gz_error(state, Z_STREAM_ERROR, c!(b"example error\0")) };
```
folkertdev, line 876:
```suggestion
        // gzclearerr should reset the eof and past flags.
        let gz_state = handle.cast::<GzState>();
        unsafe { core::ptr::addr_of_mut!((*gz_state).eof).write(true) };
        assert_eq!(unsafe { gzeof(handle) }, 0);
        unsafe { core::ptr::addr_of_mut!((*gz_state).past).write(true) };
        assert_eq!(unsafe { gzeof(handle) }, 1);
        unsafe { gzclearerr(handle) };
        assert!(!unsafe { core::ptr::addr_of!((*gz_state).eof).read() });
        assert!(!unsafe { core::ptr::addr_of!((*gz_state).past).read() });
        assert_eq!(unsafe { gzeof(handle) }, 0);

        // Set an error flag and message.
        {
            let state = (unsafe { handle.cast::<GzState>().as_mut() }).unwrap();
            unsafe { gz_error(state, Z_STREAM_ERROR, c!(b"example error\0")) };
        }
```

libz-rs-sys/src/gz.rs
```diff
@@ -535,6 +535,70 @@ pub unsafe extern "C-unwind" fn gzerror(file: gzFile, errnum: *mut c_int) -> *co
...
+    if state.mode != GzMode::GZ_READ {
+        return 0;
+    }
+
+    // Per the semantics described in the function comments above, the return value
+    // is based on state.past, rather than state.eof.
+    match state.past {
+        true => 1,
+        false => 0,
+    }
```
folkertdev, line 599:
```suggestion
    state.past as c_int
```

libz-rs-sys/src/gz.rs
folkertdev: oh and because miri is kind of slow, smaller tests are better (they can be evaluated in parallel). Does splitting this out into a test for gzeof and gzclearerror have big downsides?

libz-rs-sys/src/gz.rs
```diff
@@ -250,7 +250,10 @@ unsafe fn gzopen_help(source: Source, mode: *const c_char) -> gzFile {
             #[cfg(target_os = "linux")]
             {
                 oflag |= libc::O_LARGEFILE;
-                if cloexec {
+            }
+            if cloexec {
+                #[cfg(target_os = "linux")]
+                {
```
folkertdev, line 256: are you working around an unused error on `cloexec` here? if so, you can rename the variable to `_cloexec`, which will  not trigger that error, but still allow you to use the value in the linux-specific code.

brian-pane: Yes, it's for non-Linux targets (I ran into it on macOS). I used this approach instead of the leading underscore workaround (`_cloexec`) so that on Linux the compiler can still verify that `cloexec` is being used.

libz-rs-sys/src/gz.rs
```diff
@@ -535,6 +535,67 @@ pub unsafe extern "C-unwind" fn gzerror(file: gzFile, errnum: *mut c_int) -> *co
...
+/// # Safety
+///
+/// `file`, if non-null, must be an open file handle obtained from [`gzopen`] or [`gzdopen`].
+#[cfg_attr(feature = "export-symbols", export_name = crate::prefix!(gzeof))]
+pub unsafe extern "C-unwind" fn gzeof(file: gzFile) -> c_int {
+    // Get internal structure and check integrity
+    let Some(state) = (unsafe { file.cast::<GzState>().as_ref() }) else {
+        return 0;
+    };
+    if state.mode != GzMode::GZ_READ {
```
folkertdev, line 590: in zlib-ng, this condition is `    if (state->mode != GZ_READ && state->mode != GZ_WRITE)`

in that case you could even write something like this

```rust
match state.mode {
    GzMode::GZ_READ => state.past as _,
    GzMode::GZ_WRITE | GzMode::GZ_NONE | GzMode::GZ_APPEND => 0,
}
```

libz-rs-sys/src/gz.rs
```diff
@@ -535,6 +535,67 @@ pub unsafe extern "C-unwind" fn gzerror(file: gzFile, errnum: *mut c_int) -> *co
...
+    }
+
+    // Clear error and EOF
+    if state.mode == GzMode::GZ_READ {
+        state.eof = false;
+        state.past = false;
+    }
+
+    // Safety: we've checked state, and gz_error supports a null message argument.
+    unsafe { gz_error(state, Z_OK, ptr::null()) };
```
folkertdev, line 564: not for this PR perhaps, but could gz_error use something like this for the message?

```rust
enum Msg {
    None,
    Errcode(c_int),
    Static(&'static str),
}
```

To just reduce pointers/c strings.

libz-rs-sys/src/gz.rs
```diff
@@ -782,4 +843,62 @@ mod tests {
...
+        // gzclearerr should reset the eof and past flags.
+        unsafe { handle.cast::<GzState>().as_mut().unwrap().eof = true };
+        unsafe { handle.cast::<GzState>().as_mut().unwrap().past= true };
+        unsafe { gzclearerr(handle) };
+        assert!(! unsafe { handle.cast::<GzState>().as_ref().unwrap().eof });
+        assert!(! unsafe { handle.cast::<GzState>().as_ref().unwrap().past });
+
+        // Set an error flag and message.
+        unsafe { gz_error(handle.cast::<GzState>().as_mut().unwrap(),
+                          Z_STREAM_ERROR, c!(b"example error\0")) };
```
folkertdev, line 867: I think this needs `rustfmt`?

libz-rs-sys/src/gz.rs
```diff
@@ -782,4 +843,62 @@ mod tests {
...
+    #[test]
+    #[cfg_attr(not(target_os = "linux"), ignore = "lseek is not implemented")]
+    fn test_gzeof() {
+        // gzeof on a null file handle should return false.
+        assert_eq!(unsafe { gzeof(ptr::null_mut()) }, 0);
+
+        // Open a gzip stream with an invalid file handle. gzeof should return 0.
+        let handle = unsafe { gzdopen(-2, c!(b"r\0")) };
+        assert!(!handle.is_null());
+        assert_eq!(unsafe { gzeof(handle) }, 0);
```
folkertdev, line 893: could you add a test for when the mode is write?

libz-rs-sys/src/gz.rs
```diff
@@ -782,4 +843,62 @@ mod tests {
...
+
+    #[test]
+    #[cfg_attr(not(target_os = "linux"), ignore = "lseek is not implemented")]
+    fn test_gzclearerr() {
+        // gzclearerr on a null file handle should return quietly.
+        unsafe { gzclearerr(ptr::null_mut()) };
+
+        // Open a gzip stream with an invalid file handle. Initially, no error
+        // status should be set.
+        let handle = unsafe { gzdopen(-2, c!(b"r\0")) };
```
folkertdev, line 855: could you add a test for when the mode is write? and maybe append too?

libz-rs-sys/src/gz.rs
folkertdev: allright, that's fine then.

libz-rs-sys/src/gz.rs
brian-pane: That check in zlib-ng is redundant. It doesn't need to include the `&& state->mode != GZ_WRITE` because the next line is this:
```
    /* return end-of-file state */
    return state->mode == GZ_READ ? state->past : 0;
```

folkertdev: ah right. let's add a test for that then and we should be good

libz-rs-sys/src/gz.rs
brian-pane: Definitely! This will cover all the ways zlib-ng uses gz_error:
```
enum Msg {
    None,
    Errcode((c_int, &str))
}
```
or equivalently:
```
Option<(c_int, &str)>
```

I'm working on an update to this PR to add more test cases, and I'll include that change.

folkertdev (approved):

PR #341, opened 2025-04-13
Implement gzdirect

brian-pane:


folkertdev (commented):
Mostly looks good

- I think it'll be more convenient to use `Result` for private functions that return "success or failure"
- there are some blocks that are not covered by tests. This is mostly just fyi, and can be fixed later, but if it's easy to fix now, all the better.

brian-pane:
Thanks for reviewing, I'll post an update later today. Do you have a recommended test coverage measurement tool? I'm looking for something I can easily run locally.

folkertdev:
Yes, we use `cargo llvm-cov test --open` (or `--html` to just update the html and not open a (new) browser tab). see https://crates.io/crates/cargo-llvm-cov/0.1.13#installation for how to install that.

folkertdev:
Let's go with that for now, and then see if there is a more elegant way. It might be possible to e.g. fake a file descriptor in some cases, or expose some testing functions with an additional feature flag to `test-libz-rs-sys`.

brian-pane:
To make my new test cases work under Miri when running locally, I had to use `MIRIFLAGS="-Zmiri-disable-isolation"`. Is there an equivalent of that setting for the CI run?

libz-rs-sys/src/gz.rs
```diff
@@ -607,6 +707,212 @@ pub unsafe extern "C-unwind" fn gzeof(file: gzFile) -> c_int {
...
+    }
+
+    // No gzip header. If we were decoding gzip before, the remaining bytes
+    // are trailing garbage that can be ignored.
+    if !state.direct {
+        state.stream.avail_in = 0;
+        state.eof = true;
+        state.have = 0;
+        return 0;
+    }
```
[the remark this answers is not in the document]
brian-pane: I was able to cover the other blocks with new tests. This one got messy enough that I put in a FIXME comment for now. The cleanest way to get the `GzState` into the proper state to trigger this block is probably to finish implementing the read+decompress support, and then create a gzip file with junk content at the end.

brian-pane:
Oops, I accidentally removed the Miri isolation fix. I'm restoring it now...

libz-rs-sys/src/gz.rs
```diff
@@ -607,6 +707,228 @@ pub unsafe extern "C-unwind" fn gzeof(file: gzFile) -> c_int {
...
+        let Ok(_) = (unsafe { gz_avail(state) }) else {
+            return Err(());
+        };
+        if state.stream.avail_in == 0 {
+            return Ok(());
+        }
+    }
+
+    // Look for gzip magic bytes.
+    // Safety: `gz_avail` ensures that `next_in` points to at least `avail_in` readable bytes.
```
folkertdev, line 813: zlib-ng has a larger comment here with some helpful information

libz-rs-sys/src/gz.rs
```diff
@@ -607,6 +707,228 @@ pub unsafe extern "C-unwind" fn gzeof(file: gzFile) -> c_int {
...
+        unsafe { gz_error(state, Some((Z_MEM_ERROR, "out of memory"))) };
+        return Err(());
+    }
+
+    // Get at least the magic bytes in the input buffer.
+    if state.stream.avail_in < 2 {
+        // Safety: The caller confirmed the validity of `state`.
+        let Ok(_) = (unsafe { gz_avail(state) }) else {
+            return Err(());
+        };
```
folkertdev, line 806:
```suggestion
        let _bytes_available = gz_avail(state)?;
```

libz-rs-sys/src/gz.rs
```diff
@@ -607,6 +707,228 @@ pub unsafe extern "C-unwind" fn gzeof(file: gzFile) -> c_int {
...
+        if state.stream.avail_in == 0 {
```
folkertdev, line 807: this could use the return value from `gz_avail`. Let's do that in this case actually, so you get `if gz_avail(state)? == 0` and so on.

libz-rs-sys/src/gz.rs
```diff
@@ -607,6 +707,228 @@ pub unsafe extern "C-unwind" fn gzeof(file: gzFile) -> c_int {
...
+// `state` must have been properly initialized, e.g. by [`gzopen_help`].
+unsafe fn gz_avail(state: &mut GzState) -> Result<usize, ()> {
+    if state.err != Z_OK && state.err != Z_BUF_ERROR {
+        return Err(());
+    }
+    if !state.eof {
+        if state.stream.avail_in != 0 {
+            // Copy any remaining input to the start.
+            unsafe {
+                ptr::copy(
```
folkertdev, line 875: I suspect also here 
```suggestion
                ptr::copy_nonoverlapping(
```

libz-rs-sys/src/gz.rs
```diff
@@ -964,4 +1302,112 @@ mod tests {
...
+        // Open a gzip stream from an invalid file descriptor. gzdirect should return 1, but
+        // it should cause an error condition to be set within the stream.
+        let file = unsafe { gzdopen(-2, CString::new("r").unwrap().as_ptr()) };
+        assert!(!file.is_null());
+        assert_eq!(unsafe { gzdirect(file) }, 1);
+        let mut err = Z_OK;
+        let msg = unsafe { gzerror(file, &mut err as *mut c_int) };
+        assert!(!msg.is_null());
+        assert_eq!(err, Z_ERRNO);
+        assert_eq!(unsafe { gzclose(file) }, Z_ERRNO);
```
folkertdev, line 1321: this block can go into `test-libz-rs-sys` right?

libz-rs-sys/src/gz.rs
```diff
@@ -478,20 +512,86 @@ unsafe fn deallocate_cstr(s: *mut c_char) {
...
+// `state` must have been properly initialized, e.g. by [`gzopen_help`].
+unsafe fn gzclose_r(state: &mut GzState) -> c_int {
+    if state.mode != GzMode::GZ_READ {
+        return Z_STREAM_ERROR;
+    }
+
+    // Free buffers and close the file
+    if state.in_size != 0 {
+        // Safety: state.stream was properly initialized as a z_stream in gzopen_help.
+        unsafe { inflateEnd(&mut state.stream as *mut z_stream) };
```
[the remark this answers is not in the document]
brian-pane: I'll add a comment and post an update later today. The short answer is: the Rust versions of `gzclose_r` and `gzclose_w` receive a reference to the `GzState`, instead of a pointer. We can't deallocate that state inside those functions while the reference is in scope.

libz-rs-sys/src/gz.rs
brian-pane: In this case, we're shifting data from the middle of a buffer to the front, so it may overlap.

libz-rs-sys/src/gz.rs
```diff
@@ -478,20 +512,86 @@ unsafe fn deallocate_cstr(s: *mut c_char) {
 /// This function may be called at most once for any file handle.
 #[cfg_attr(feature = "export-symbols", export_name = crate::prefix!(gzclose))]
 pub unsafe extern "C-unwind" fn gzclose(file: gzFile) -> c_int {
-    let Some(state) = (unsafe { file.cast::<GzState>().as_mut() }) else {
-        return Z_STREAM_ERROR;
+    let ret = match unsafe { file.cast::<GzState>().as_mut() } {
+        Some(state) => match state.mode {
+            GzMode::GZ_READ => unsafe { gzclose_r(state) },
+            _ => gzclose_w(state),
```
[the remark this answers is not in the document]
brian-pane: They key this about this version, compared to the previous version, is that it puts the reference `state` within a nested scope. That reference's lifetime needs to end before the call to `free_state` or Miri will (correctly) raise an error when `free_state` deallocates the memory underlying a life reference. I'll add a comment about it.

libz-rs-sys/src/gz.rs
```diff
@@ -607,6 +707,228 @@ pub unsafe extern "C-unwind" fn gzeof(file: gzFile) -> c_int {
...
+// # Returns
+//
+// * `Ok` on success
+// * `Err` on failure
+//
+// # Safety
+//
+// `state` must have been properly initialized, e.g. by [`gzopen_help`].
+unsafe fn gz_look(state: &mut GzState) -> Result<(), ()> {
+    // Allocate buffers if needed.
```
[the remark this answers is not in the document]
brian-pane: zlib-ng uses a single `size` field to store of the input and output buffer sizes, except that it's not really the buffer size, because of this:
```
    unsigned size;          /* buffer size, zero if not allocated yet */
    unsigned want;          /* requested buffer size, default is GZBUFSIZE */
    unsigned char *in;      /* input buffer (double-sized when writing) */
    unsigned char *out;     /* output buffer (double-sized when reading) */
```

One of the input or output buffer will be twice `state->size`, depending on the mode. In the Rust version, where we have to specify the exact buffer size when deallocating, I used `in_size` and `out_size` fields to reduce the potential sources of error.

Since the check for `state->size == 0` in `gz_look` is deciding whether it needs to allocate the I/O buffers, I used an equivalent check: `state.input.is_null()`. The check could alternately be `state.in_size == 0`.

libz-rs-sys/src/gz.rs
```diff
@@ -1353,8 +1349,13 @@ mod tests {
             assert_eq!(state.stream.avail_in as usize, expected_avail);
         };
         assert_eq!(unsafe { gzclose(file) }, Z_OK);
+    }
 
-        // Open a non-gzip file for reading. gzdirect should return 1.
+    // Open a non-gzip file for reading. gzdirect should return 1.
+    #[test]
+    #[cfg_attr(not(target_os = "linux"), ignore = "lseek is not implemented")]
+    // Open a gzip file for reading. gzdirect should return 0.
```
folkertdev, line 1357: something weird happed with the comments here.

libz-rs-sys/src/gz.rs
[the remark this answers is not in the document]
folkertdev: Ah, nice. Where would we be without Miri.

libz-rs-sys/src/gz.rs
folkertdev: Maybe add a note then that they may overlap

PR #344, opened 2025-04-17
Fuzz coverage error paths

folkertdev:
This makes three changes to the uncompress fuzzer

- rename `uncompress2` to `uncompress` (the old `uncompress` fuzzer was removed)
- when gathering code coverage, keep invalid inputs in the corpus so that code coverage shows what error paths we cover.
- feed the input in chunks of 64 bytes, to exercise more of the streaming logic

With those changes we get much better coverage of `inflate.rs`:

https://app.codecov.io/github/trifectatechfoundation/zlib-rs/commit/291a345601f0e817b1fb2a385e3df400e576662b/blob/zlib-rs/src/inflate.rs?flags%5B0%5D=fuzz-decompress&dropdown=coverage

cc @inahga


fuzz/fuzz_targets/uncompress.rs
```diff
@@ -1,54 +1,147 @@
...
-
-    let error = unsafe {
-        libz_ng_sys::compress(
-            deflated.as_mut_ptr().cast(),
-            &mut length,
-            data.as_ptr().cast(),
-            data.len(),
-        )
+    // Small enough to hit interesting cases, but large enough to hit the fast path
+    let chunk_size = 64;
```
bjorn3, line 81: Would varying the chunk size make sense? Especially non-power-of-two chunk sizes seem like they could be useful for finding edge-cases.

folkertdev: It might, but I don't see how we could do that in a deterministic way and continue to use the compression corpus.

bjorn3: You could take the chunk size from the fuzz input, right? So for example interpreting the fuzz input as series of 1 byte chunk size + n bytes chunk data.

folkertdev: Would that not mess with how effective the corpus is?

bjorn3: You could reformat the existing corpus into this format, right? In any case I guess this is better left as future improvement.

bjorn3 (approved):

PR #345, opened 2025-04-17
Implement gzread and gzbuffer

brian-pane:


brian-pane:
The Miri run in the CI doesn't report the error I get when I run it locally:
```
test gz::gread_gzip ... error: Undefined Behavior: constructing invalid value at .value[6]: encountered uninitialized memory, but expected an integer
    --> /home/brianp/code/zlib-rs/zlib-rs/src/inflate/writer.rs:299:5
     |
299  |     core::ptr::read_unaligned(from.cast::<[u8; N]>())
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ constructing invalid value at .value[6]: encountered uninitialized memory, but expected an integer
     |
     = note: BACKTRACE on thread `gz::gread_gzip`:
     = note: inside `zlib_rs::inflate::writer::load_chunk::<16>` at /home/brianp/code/zlib-rs/zlib-rs/src/inflate/writer.rs:299:5: 299:54
note: inside `zlib_rs::inflate::writer::Writer::<'_>::copy_chunk_unchecked::<16>`
    --> /home/brianp/code/zlib-rs/zlib-rs/src/inflate/writer.rs:278:21
     |
278  |         let chunk = load_chunk::<N>(src);
     |                     ^^^^^^^^^^^^^^^^^^^^
note: inside `zlib_rs::inflate::writer::Writer::<'_>::copy_chunked_within::<16>`
    --> /home/brianp/code/zlib-rs/zlib-rs/src/inflate/writer.rs:259:22
     |
259  | ...safe { Self::copy_chunk_unchecked::<N>(ptr.add(start), ptr.add(current), length) }
     |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside `zlib_rs::inflate::writer::Writer::<'_>::copy_match_help::<16>`
    --> /home/brianp/code/zlib-rs/zlib-rs/src/inflate/writer.rs:242:13
     |
242  |             Self::copy_chunked_within::<N>(buf, capacity, current, offset_from_end, length)
     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside `zlib_rs::inflate::writer::Writer::<'_>::copy_match_runtime_dispatch`
    --> /home/brianp/code/zlib-rs/zlib-rs/src/inflate/writer.rs:195:20
     |
195  |             return self.copy_match_help::<16>(offset_from_end, length);
     |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside `zlib_rs::inflate::writer::Writer::<'_>::copy_match_with_features::<0>`
    --> /home/brianp/code/zlib-rs/zlib-rs/src/inflate/writer.rs:176:18
     |
176  |             _ => self.copy_match_runtime_dispatch(offset_from_end, length),
     |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside `zlib_rs::inflate::writer::Writer::<'_>::copy_match`
    --> /home/brianp/code/zlib-rs/zlib-rs/src/inflate/writer.rs:164:9
     |
164  |         self.copy_match_with_features::<{ CpuFeatures::NONE }>(offset_from_end, length)
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside `zlib_rs::inflate::State::<'_>::len_and_friends`
    --> /home/brianp/code/zlib-rs/zlib-rs/src/inflate.rs:819:29
     |
819  | ...                   writer.copy_match(self.offset, copy);
     |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside `zlib_rs::inflate::State::<'_>::dispatch`
    --> /home/brianp/code/zlib-rs/zlib-rs/src/inflate.rs:1381:35
     |
1381 |                         let val = self.len_and_friends();
     |                                   ^^^^^^^^^^^^^^^^^^^^^^
note: inside `zlib_rs::inflate::inflate`
    --> /home/brianp/code/zlib-rs/zlib-rs/src/inflate.rs:2242:19
     |
2242 |     let mut err = state.dispatch();
     |                   ^^^^^^^^^^^^^^^^
note: inside `libz_rs_sys::inflate`
    --> /home/brianp/code/zlib-rs/libz-rs-sys/src/lib.rs:362:9
     |
362  |         zlib_rs::inflate::inflate(stream, flush) as _
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside `libz_rs_sys::gz::gz_decomp`
    --> /home/brianp/code/zlib-rs/libz-rs-sys/src/gz.rs:1252:24
     |
1252 |         match unsafe { inflate(&mut state.stream, Z_NO_FLUSH) } {
     |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside `libz_rs_sys::gz::gz_fetch`
    --> /home/brianp/code/zlib-rs/libz-rs-sys/src/gz.rs:1208:26
     |
1208 |                 unsafe { gz_decomp(state) }?;
     |                          ^^^^^^^^^^^^^^^^
note: inside `libz_rs_sys::gz::gz_read`
    --> /home/brianp/code/zlib-rs/libz-rs-sys/src/gz.rs:943:25
     |
943  |             if unsafe { gz_fetch(state) }.is_err() {
     |                         ^^^^^^^^^^^^^^^
note: inside `libz_rs_sys::gzread`
    --> /home/brianp/code/zlib-rs/libz-rs-sys/src/gz.rs:877:24
     |
877  |     let got = unsafe { gz_read(state, buf.cast::<u8>(), len as usize) };
     |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside `gz::gread_gzip`
    --> test-libz-rs-sys/src/gz.rs:300:18
     |
300  |         unsafe { gzread(file, buf.as_mut_ptr().cast::<c_void>(), 10) },
     |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside closure
    --> test-libz-rs-sys/src/gz.rs:265:16
     |
264  | #[test]
     | ------- in this procedural macro expansion
265  | fn gread_gzip() {
```

folkertdev:
What command do you run?

folkertdev:
I think the miri error is actually a soundness bug in the inflate code, where some uninitialized memory is read. We should not be doing that of course, but I don't think it's dangerous in this case. I'll fix that separately.

brian-pane:
My local Miri command is `MIRIFLAGS="-Zmiri-disable-isolation" cargo +nightly miri test gz::gread_gzip` ... and if I patch in #346 it fixed the error message I was getting.

brian-pane:
I'm not sure what's happening in the failed x86_64-pc-windows-gnu test. I'll try to set up a local Windows dev environment tomorrow to repro it.

brian-pane:
I finally got a Rust `x86_64-pc-windows-gnu` build working locally (under MSYS2 UCRT64), but all the tests pass on my system. I'm not sure how to diagnose the CI Windows test error.

brian-pane:
I got the test error to repro in a local `x86_64-pc-windows-gnu` environment. I'll debug there to try to fix it.

folkertdev:
I can reproduce the problem by using `wine` to evaluate the test. I've added the following to my local `.cargo/config.toml` file:

```toml
testw = "test --target x86_64-pc-windows-gnu"

[target.'cfg(windows)']
runner = "wine"
```

The I can run these commands to gather the stdout, and add `dbg!`s to try and figure out where the two implementations differ

```
> cargo test --features="gz" gread_gzip -- --nocapture 2&> /tmp/old.txt
> cargo testw --features="gz" gread_gzip -- --nocapture 2&> /tmp/new.txt
```

I use `meld` to quickly see where the two outputs diverge. The first difference between the functions is that a read on windows here returns `0`

https://github.com/trifectatechfoundation/zlib-rs/blob/1e3d9ec663f00a692ed9f91021d3088191812e14/libz-rs-sys/src/gz.rs#L926-L930

brian-pane:
Yes, on Windows I see libc read returning 128 bytes on the first five reads, then 22 bytes, then zero.

brian-pane:
The latest change that I just pushed adds a little test case called `libc_read` in `test-libz-rs-sys` that bypasses all the `gz` code and iterates on the libc read call to read the test file. That test case only manages to read 662 bytes from the file on Windows, which is the same number that the `gread_gzip` test reads before a read returns zero. That's consistent, at least, but it's surprising. Does libc read work differently on Windows than on Posix?

folkertdev:
ah yes, you need to add the `libc::O_BINARY` flag on windows to the `libc::open` so that the file is opened in binary mode

brian-pane:
Thanks! `O_BINARY` fixed the issue. Now the test case that reads a text file is failing - but only in the Windows CI run, not when I run it locally on Windows. In the CI `x86_64-pc-windows-gnu` test failure, the number of bytes read from the file matches what I get if I pass the file through `unix2dos`. My guess is that the file gets translated to DOS-style newlines when checked out by `git` in the Windows test environment. I'll rework my test case to validate the number of bytes read against the file size returned by `stat`, rather than hard-coding a check for what the file size is on Linux.

libz-rs-sys/src/gz.rs
```diff
@@ -753,6 +812,188 @@ pub unsafe extern "C-unwind" fn gzdirect(file: gzFile) -> c_int {
...
+        }
+    }
+     */
+
+    // Loop until we get enough bytes or reach the end of the file.
+    let mut got = 0;
+    let mut count = 0;
+    loop {
+        count += 1;
+        assert_ne!(count, 100);
```
folkertdev, line 923: Is this left from debugging?

libz-rs-sys/src/gz.rs
```diff
@@ -753,6 +812,188 @@ pub unsafe extern "C-unwind" fn gzdirect(file: gzFile) -> c_int {
...
+            // For small len or a new stream, load more data from the file descriptor into
+            // the output buffer. Note: If we haven't scanned the file header yet, gz_fetch
+            // will read the header and determine whether to use decompression or direct read.
+            if unsafe { gz_fetch(state) }.is_err() {
+                return 0;
+            }
+            assert_ne!(state.have, 0);
+            continue; // Now that we've tried reading, we can try to copy from the output buffer.
+                      // The copy above assures that we will leave with space in the
+                      // output buffer, allowing at least one gzungetc() to succeed.
```
folkertdev, line 959: this reads nicer if you put the comment above the `continue` (and maybe add an empty line after the `assert_ne`

folkertdev:
2 final minor things, otherwise this looks good!

libz-rs-sys/src/gz.rs
brian-pane: Oops! Leftover from debugging. Thanks for catching that!

PR #346, opened 2025-04-17
Copy chunks as `MaybeUninit<u8>`

folkertdev:
When the output buffer contained uninitialized memory, MIRI reported an error in the `copy_chunk_unchecked` function. That is unsound, though I don't see any way that could have been observed or exploited.

The bounds checks are all valid as far as I can see, so I think it's OK to just read/write the memory without assuming that it is initialized. Let's see if this has a performance impact though.


folkertdev:
perf looks ok to me https://github.com/trifectatechfoundation/zlib-rs/actions/runs/14524839981

bjorn3 (approved):

PR #347, opened 2025-04-21
Implement gzwrite and gzflush

brian-pane:


libz-rs-sys/src/gz.rs
```diff
@@ -1296,6 +1303,340 @@ unsafe fn gz_decomp(state: &mut GzState) -> Result<(), ()> {
...
+fn gz_init(state: &mut GzState) -> Result<(), ()> {
+    // Allocate input buffer (double size for gzprintf).
+    state.in_size = state.want * 2;
+    let Some(input) = ALLOCATOR.allocate_slice_raw::<u8>(state.in_size) else {
+        // Safety: The caller confirmed the validity of state.
+        unsafe { gz_error(state, Some((Z_MEM_ERROR, "out of memory"))) };
+        return Err(());
+    };
+    state.input = input.as_ptr();
+    // Note: zlib-ng fills the input buffer with zeroes here, but it's unneeded.
```
folkertdev, line 1449: Are you sure there won't be any issues with reading uninitialized memory?

libz-rs-sys/src/gz.rs
```diff
@@ -1296,6 +1303,340 @@ unsafe fn gz_decomp(state: &mut GzState) -> Result<(), ()> {
...
+    if state.direct {
+        let got = unsafe {
+            libc::write(
+                state.fd,
+                state.stream.next_in.cast::<c_void>(),
+                state.stream.avail_in as _,
+            )
+        };
+        if got < 0 || got as c_uint != state.stream.avail_in {
+            unsafe { gz_error(state, Some((Z_ERRNO, "write error"))) };
```
folkertdev, line 1526: The C version uses `zstrerror` here. You just assume that the error will be `"write error"` in this case?

libz-rs-sys/src/gz.rs
```diff
@@ -1296,6 +1303,340 @@ unsafe fn gz_decomp(state: &mut GzState) -> Result<(), ()> {
...
+        // doing Z_FINISH then don't write until we get to Z_STREAM_END.
+        if state.stream.avail_out == 0
+            || (flush != Z_NO_FLUSH && (flush != Z_FINISH || ret == Z_STREAM_END))
+        {
+            // Safety: Within this gz module, `state.stream.next` and `state.stream.next_out`
+            // always point within the same allocated object, `state.stream.output`.
+            let have = unsafe { state.stream.next_out.offset_from(state.next) };
+            if have != 0 {
+                let ret = unsafe { libc::write(state.fd, state.next.cast::<c_void>(), have as _) };
+                if ret != have as _ {
```
folkertdev, line 1558: the C code has a check for `ret < 0` as well here. I don't think `have` is guaranteed to be positive here?

libz-rs-sys/src/gz.rs
```diff
@@ -1296,6 +1303,340 @@ unsafe fn gz_decomp(state: &mut GzState) -> Result<(), ()> {
...
+        let mut have = state.stream.avail_out;
+        ret = unsafe { deflate(&mut state.stream, flush) };
+        if ret == Z_STREAM_ERROR {
+            unsafe {
+                gz_error(
+                    state,
+                    Some((Z_STREAM_ERROR, "internal error: deflate stream corrupt")),
+                )
+            };
+            return Err(());
```
folkertdev, line 1580:
```suggestion
            let msg = "internal error: deflate stream corrupt";
            unsafe { gz_error(state, Some((Z_STREAM_ERROR, msg))) };
            return Err(());
```

These `gz_error` calls are taking up a lot of lines, so something like the above is totally OK to reduce the impact a bit.

libz-rs-sys/src/gz.rs
```diff
@@ -1296,6 +1303,340 @@ unsafe fn gz_decomp(state: &mut GzState) -> Result<(), ()> {
...
+                state
+                    .stream
+                    .next_in
+                    .add(state.stream.avail_in as usize)
+                    .offset_from(state.input)
+            } as usize;
+            let copy = cmp::min(state.in_size.saturating_sub(have), len as usize);
+            // Safety: The caller is responsible for ensuring that buf points to at least len readable
+            // bytes, and copy is <= len.
+            unsafe { ptr::copy(buf, state.input.add(have).cast::<c_void>(), copy) };
```
folkertdev, line 1395: `buf` is only readable memory (per the safety comment), so this copy should/could be non-overlapping then right?

libz-rs-sys/src/gz.rs
```diff
@@ -1296,6 +1303,340 @@ unsafe fn gz_decomp(state: &mut GzState) -> Result<(), ()> {
...
+    } else {
+        // Consume any data left in the input buffer.
+        if state.stream.avail_in != 0 && gz_comp(state, Z_NO_FLUSH).is_err() {
+            return 0;
+        }
+
+        // Directly compress user buffer to file.
+        state.stream.next_in = buf.cast::<_>();
+        loop {
+            let n = cmp::min(len, c_uint::MAX);
```
folkertdev, line 1416: This seems kind of pointless? In the original code, `len` is a `size_t`, which I think can be 64-bit when `c_uint` is only 32-bit.

libz-rs-sys/src/gz.rs
```diff
@@ -1296,6 +1303,340 @@ unsafe fn gz_decomp(state: &mut GzState) -> Result<(), ()> {
...
+// # Returns
+// - The number of uncompress bytes written, on success.
+// - Or 0 in case of error.
+//
+// # Safety
+//
+/// - `state` must have been properly initialized, e.g. by [`gzopen_help`].
+/// - `buf` must point to at least `len` bytes of readable memory.
+/// - The caller must ensure that `len` fits in a `c_int`.
+unsafe fn gz_write(state: &mut GzState, mut buf: *const c_void, mut len: c_uint) -> c_int {
```
folkertdev, line 1356: the C code uses `size_t len`, that's not equivalent to `c_uint` right?

libz-rs-sys/src/gz.rs
brian-pane: The C API exposes `state.next` as part of `gzFile_s`, so it's possible for C code using this library to do something like:
```
void* buf = file->next;
unsigned len = file->have;
gzwrite(file, buf, len);
```

I think `next` can only ever point into `state.output` and not `state.input`, so even a call like that will lead to a non-overlapping copy here, but I need to look through all the code paths to make sure.

libz-rs-sys/src/gz.rs
brian-pane: Right, I'll change this to `usize`.

libz-rs-sys/src/gz.rs
brian-pane: I'll change the `len` parameter in this function to `usize` for consistency with the `size_t` in the original, which will make this check relevant.

folkertdev: Actually when `len` is a `usize` this operation is actually meaningful.

libz-rs-sys/src/gz.rs
brian-pane: Interestingly, there's a problem with using `usize`. As far as I can tell, a C `unsigned int` is allowed to be bigger than a C pointer, in case someone ever builds a computer with, for example, a 32-bit address space and a 128-bit ALU. So this code path:
```
pub unsafe extern "C-unwind" fn gzwrite(file: gzFile, buf: *const c_void, len: c_uint) -> c_int {
    ...
    gz_write(state, buf, len as usize)
}
```
could lead to a silent truncation of the `len` value when casting to `usize`. The safety preconditions I've documented for `gzwrite` include "`buf` must point to at least `len` bytes of readable memory," which ensures that `len` will fit in a `usize`. But I'll also use a checked cast here for extra future-proofing.

libz-rs-sys/src/gz.rs
brian-pane: The `gz` code only reads parts of `input` that have been written (confirmed by Miri). As long as the `inflate` and `deflate` code don't read past the filled part of `input`, we won't need to zero out the memory.

libz-rs-sys/src/gz.rs
brian-pane: Yes, I left a FIXME comment about implementing `zstrerror` in the corresponding part of the read path, but I forgot to add it here.

Implementing `zstrerror` is probably big enough to live in a separate PR. zlib-ng aliases it to the C `strerror` function, which [isn't necessarily thread-safe](https://pubs.opengroup.org/onlinepubs/9699919799/functions/strerror.html). Probably we should use `strerror_r`, although that isn't available on all platforms (a scan of the Rust `libc` source suggests it's missing on Win32, for example). For now I'll add a FIXME comment here.

libz-rs-sys/src/gz.rs
brian-pane: `have` is cast to an unsigned value (implicitly in the C code, and with `as _` here) when passed as the length argument of `write`, so if it's negative we may crash in `write` when it reads far past the end of the buffer. The C implementation relies on this subtraction being positive:
```
have = (unsigned)(strm->next_out - state->x.next)
```
For the Rust version, I'll check for `have < 0` and return an error prior to the `write`.

libz-rs-sys/src/gz.rs
folkertdev: They won't read past the range that you give it. assuming you don't give it uninitialized input bytes we should be good then.

libz-rs-sys/src/gz.rs
```diff
@@ -866,13 +873,9 @@ pub unsafe extern "C-unwind" fn gzread(file: gzFile, buf: *mut c_void, len: c_ui
 
     // Check that the requested number of bytes can be represented by the return type.
     if c_int::try_from(len).is_err() {
+        const MSG: &str = "request does not fit in an int";
```
folkertdev, line 876: just as a note, `const MSG: &str = "some literal"` is the same as `let msg = "some literal"`. String literals always get a static lifetime and are stored in the directly in the binary, just like constants.

folkertdev (approved):
LGTM then

PR #348, opened 2025-04-21
Make gzclose_r and gzclose_w part of the public API

brian-pane:


folkertdev:
These symbols are not exported by `libz.so`, so why would we make them part of the public API?

brian-pane:
They are part of the public API in zlib-ng, https://github.com/zlib-ng/zlib-ng/blob/2.2.4/zlib.h.in#L1624-L1625

libz-rs-sys/src/gz.rs
```diff
@@ -521,89 +520,127 @@ unsafe fn deallocate_cstr(s: *mut c_char) {
...
-        },
-        None => Z_STREAM_ERROR,
+    let mode = match unsafe { file.cast::<GzState>().as_ref() } {
+        // Note: It is important that this `state` reference lives in a nested scope
+        // because its lifetime must not overlap the calls to `gzclose_r`/`gzclose_w` below,
+        // which free the referenced memory.
+        Some(state) => state.mode,
+        None => {
+            return Z_STREAM_ERROR;
+        }
```
folkertdev, line 530:
```suggestion
        None => return Z_STREAM_ERROR,
```

libz-rs-sys/src/gz.rs
```diff
@@ -521,89 +520,127 @@ unsafe fn deallocate_cstr(s: *mut c_char) {
...
-    // Note: In zlib-ng, the gzclose_r and gzclose_w functions deallocate the state
-    // object themselves. Here, since those functions take references to the state, the
-    // deallocation cannot safely happen until the references are out of scope. Therefore
-    // the free_state call handles the deallocation using the raw pointer to the state.
-    unsafe { free_state(file.cast::<GzState>()) };
-
-    ret
+    match mode {
+        GzMode::GZ_READ => unsafe { gzclose_r(file) },
+        _ => unsafe { gzclose_w(file) },
```
folkertdev, line 535: may as well make this exhaustive right? 

```suggestion
        GzMode::GZ_WRITE | GzMode::GZ_APPEND | GzMode::GZ_NONE => unsafe { gzclose_w(file) },
```

libz-rs-sys/src/gz.rs
```diff
@@ -521,89 +520,127 @@ unsafe fn deallocate_cstr(s: *mut c_char) {
...
+        }
     };
-    match unsafe { libc::close(state.fd) } {
-        0 => err,
-        _ => Z_ERRNO,
-    }
+
+    // Now that the `state` reference has gone out of scope, we can safely
+    // delete the underlying allocation.
+    unsafe { free_state(file.cast::<GzState>()) };
```
folkertdev, line 584: just to be sure: miri actively complains if you don't put the above in a block? In some cases rust can see that a reference is no longer in use and then more things are allowed.

folkertdev (commented):
Some nits/questions.

In general, are there any tests that can be moved from `libz-rs-sys` to `test-libz-rs-sys` now that these functions are public?

libz-rs-sys/src/gz.rs
brian-pane: Miri doesn't complain, so I'll un-nest the block and add a safety comment on the `free_state` to document that the reference can't be used beyond that point.

brian-pane:
I just checked the existing tests, and they all exercise these functions indirectly via the `gzclose` that was already public, so there aren't any test cases that can move from `libz-rs-sys` to `test-libz-rs-sys`.

folkertdev (approved):

PR #349, opened 2025-04-23
Implement gzoffset and gztell

brian-pane:


brian-pane:
I'll debug the remaining error in a local Windows environment.

PR #351, opened 2025-04-23
Implement gzputc and gzputs

brian-pane:


brian-pane:
There are a couple of test coverage gaps that I want to fix, and then I'll move this PR out of draft state.

libz-rs-sys/src/gz.rs
```diff
@@ -1738,6 +1738,111 @@ pub unsafe extern "C-unwind" fn gzoffset(file: gzFile) -> z_off_t {
...
+        }
+    }
+
+    // No room in buffer or not initialized, use gz_write.
+    let buf = [c as u8];
+    // Safety: We have confirmed that `state` is valid, and `buf` contains 1 readable byte of data.
+    if unsafe { gz_write(state, buf.as_ptr().cast::<c_void>(), 1) } != 1 {
+        return -1;
+    }
+    c & 0xff
```
folkertdev, line 1798:
```suggestion
    match unsafe { gz_write(state, buf.as_ptr().cast::<c_void>(), 1) } {
        1 => c & 0xff,
        _ => -1,
    }
```

libz-rs-sys/src/gz.rs
```diff
@@ -1738,6 +1738,111 @@ pub unsafe extern "C-unwind" fn gzoffset(file: gzFile) -> z_off_t {
...
+            )
+        };
+        return -1;
+    }
+    let put = unsafe { gz_write(state, s.cast::<c_void>(), len) };
+    if put < len as _ {
+        -1
+    } else {
+        len as _
+    }
```
folkertdev, line 1843:
```suggestion
    match put.cmp(&len) {
        Ordering::Less => -1,
        Ordering::Equal | Ordering::Greater => len as _,
    }
```

libz-rs-sys/src/gz.rs
```diff
@@ -1738,6 +1738,111 @@ pub unsafe extern "C-unwind" fn gzoffset(file: gzFile) -> z_off_t {
...
+        return -1;
+    }
+
+    // Write string.
+    let len = unsafe { libc::strlen(s) };
+    if c_int::try_from(len).is_err() {
+        unsafe {
+            gz_error(
+                state,
+                Some((Z_STREAM_ERROR, "string length does not fit in int")),
```
folkertdev, line 1833: using `let msg = "..."` should reduce the linecount here

libz-rs-sys/src/gz.rs
```diff
@@ -1738,6 +1738,111 @@ pub unsafe extern "C-unwind" fn gzoffset(file: gzFile) -> z_off_t {
...
+        if state.stream.avail_in == 0 {
+            state.stream.next_in = state.input;
+        }
+        let have = unsafe {
+            state
+                .stream
+                .next_in
+                .add(state.stream.avail_in as usize)
+                .offset_from(state.input)
+        } as usize;
```
folkertdev, line 1782: We already have this bit of code somewhere else right? Maybe this can be a method on `state`?

brian-pane: I just pushed an update where this is encapsulated in a function. I made a small change to `gz_write` to simplify the safety conditions for this function. Let me know if the safety comments make sense.

PR #352, opened 2025-04-24
Implement gzgetc and gzgets

brian-pane:


brian-pane:
The coverage report shows no calls to `gzgetc` or `gzgets`, but the `cargo nextest` output from the "Build and test" CI runs shows the test cases being run. I'm not sure what's happening there.

There also is an error in the wasm32 CI build, but I don't know enough about the environment to understand it:
```
  Error: failed to compile: wasm[0]::function[56]::_ZN16test_libz_rs_sys2gz12gzgetc_basic17hebe6d9ec78d726bcE

  Caused by:
      0: WebAssembly translation error
      1: Invalid input WebAssembly code at offset 47607: legacy exceptions support is not enabled
```

folkertdev:
> The coverage report shows no calls to gzgetc or gzgets, but the cargo nextest output from the "Build and test" CI runs shows the test cases being run. I'm not sure what's happening there.

I'm not entirely sure either, you can see the correct codecov by using:

```
cargo llvm-cov nextest --features="gz" -p test-libz-rs-sys
```

bjorn3:
According to the error for some reason part of the code gets compiled with wasm exceptions enabled.

libz-rs-sys/src/gz.rs
```diff
@@ -1861,6 +1863,186 @@ pub unsafe extern "C-unwind" fn gzputs(file: gzFile, s: *const c_char) -> c_int
...
+        return ret as _;
+    }
+
+    // Nothing there -- try gz_read.
+    let mut buf = [0u8; 1];
+    // Safety: `buf` is big enough to hold `len = 1` bytes.
+    match unsafe { gz_read(state, buf.as_mut_ptr(), 1) } {
+        1 => buf[0] as _,
+        _ => -1,
+    }
```
folkertdev, line 1910:
```suggestion
    let mut c = 0u8;
    // Safety: `buf` is big enough to hold `len = 1` bytes.
    match unsafe { gz_read(state, core::slice::from_mut(&mut c).as_mut_ptr(), 1) } {
        1 => c_int::from(c),
        _ => -1,
    }
```

libz-rs-sys/src/gz.rs
```diff
@@ -1861,6 +1863,186 @@ pub unsafe extern "C-unwind" fn gzputs(file: gzFile, s: *const c_char) -> c_int
...
+        state.have -= 1;
+        state.pos += 1;
+        // Safety: Since `state.have` is at least 1, `state.next` points to at least
+        // one readable byte within `state.output`.
+        let ret = unsafe { *state.next };
+        // Safety: Since `state.have` is at least 1, the byte between `state.next` and
+        // `state.next + 1` is within the bounds of the `state.output` buffer, as required
+        // by the pointer `add` method.
+        state.next = unsafe { state.next.add(1) };
+        return ret as _;
```
folkertdev, line 1901:
```suggestion
        return c_int::from(ret);
```

using `from` is a signal that no truncation is happening. `From` for the builtin numeric types is only implemented when the cast is lossless.

folkertdev:
> According to the error for some reason part of the code gets compiled with wasm exceptions enabled.

But if you look at the fix, I don't think that is true? I suspect it's a wasmtime bug that just shows up with that error message because it found some bytes it did not understand.

folkertdev:
here is a reproduction

```rust
pub extern "C-unwind" fn gzgetc() -> i32 {
    42
}

pub extern "C-unwind" fn gzgetc_() -> i32 {
    gzgetc()
}

fn main() {
    for gzgetc_fn in [gzgetc, gzgetc_] {
        // gzgetc on a null file handle should return -1.
        assert_eq!(gzgetc_fn(), -1);
    }
}
```

The use of function pointers in combination with `"C-unwind"` is problematic. I think in this case rust is not able to optimize away the panic that `C-unwind` would throw. And if I remember correctly that panic is a little special, because of course normal panics work fine for wasmtime.

bjorn3:
`extern "C-unwind"` shouldn't use exception handling if the exception-handling feature is disabled as it is by default for wasm32-wasip1. I would call this a rustc bug. Opened https://github.com/rust-lang/rust/issues/140293.

test-libz-rs-sys/src/gz.rs
```diff
@@ -1022,6 +1022,154 @@ fn gzputs_error() {
     assert_eq!(unsafe { gzclose(file) }, Z_ERRNO);
 }
 
+#[test]
+fn gzgetc_basic() {
+    // Read data from a gzip file one byte at a time using gzgetc, and verify that
+    // the expected content is returned.
+    for gzgetc_fn in [|x| unsafe { gzgetc(x) }, |x| unsafe { gzgetc_(x) }] {
```
folkertdev, line 1029: @brian-pane can you add a fixme comment here (and below where we do something similar) linking to https://github.com/rust-lang/rust/issues/140293?

folkertdev (approved):

PR #356, opened 2025-05-01
Implement gzsetparams

brian-pane:


libz-rs-sys/src/gz.rs
```diff
@@ -2269,6 +2269,64 @@ pub unsafe extern "C-unwind" fn gzgets(file: gzFile, buf: *mut c_char, len: c_in
...
+/// - `Z_MEM_ERROR` if there is a memory allocation error.
+///
+/// # Safety
+///
+/// - `file`, if non-null, must be an open file handle obtained from [`gzopen`] or [`gzdopen`].
+#[cfg_attr(feature = "export-symbols", export_name = crate::prefix!(gzsetparams))]
+pub unsafe extern "C-unwind" fn gzsetparams(file: gzFile, level: c_int, strategy: c_int) -> c_int {
+    let Ok(strategy) = Strategy::try_from(strategy) else {
+        return Z_STREAM_ERROR;
+    };
```
brian-pane, line 2290: This check isn't in zlib-ng, but I needed it for the Rust version because we use a `Strategy` enum to hold the strategy internally.  I return `Z_STREAM_ERROR` here to for consistency with the Rust version of `deflateParams`.

libz-rs-sys/src/gz.rs
```diff
@@ -2269,6 +2269,64 @@ pub unsafe extern "C-unwind" fn gzgets(file: gzFile, buf: *mut c_char, len: c_in
     }
 }
 
+/// Dynamically update the compression level and strategy for `file`. See the
+/// description of [`deflateInit2`] for the meaning of these parameters. Previously
+/// provided data is flushed before applying the parameter changes.
+///
+/// # Returns
+///
+/// - `Z_OK` on success.
```
folkertdev, line 2278:
```suggestion
/// - [`Z_OK`] on success.
```

That turns these into clickable links in the docs. Same for the others.

brian-pane:
@folkertdev this isn't an issue for this PR, but do you have a preferred way to implement C FFI varargs? One of the few gz functions left to implement is:
```
Z_EXTERN int Z_EXPORTVA gzprintf(gzFile file, const char *format, ...);
```
There's a `va_list` in core:ffi that I could use for this, but it's still experimental in the current Rust version.

libz-rs-sys/src/gz.rs
```diff
@@ -2269,6 +2269,64 @@ pub unsafe extern "C-unwind" fn gzgets(file: gzFile, buf: *mut c_char, len: c_in
...
+        return Z_STREAM_ERROR;
+    };
+
+    // Check that we're writing and that there's no error.
+    if state.mode != GzMode::GZ_WRITE || state.err != Z_OK || state.direct {
+        return Z_STREAM_ERROR;
+    }
+
+    // If no change is requested, then do nothing.
+    if level == state.level as c_int && strategy == state.strategy {
```
folkertdev, line 2301: this can be `c_int::from(state.level)` (communicates that no truncation can happen)

libz-rs-sys/src/gz.rs
```diff
@@ -2269,6 +2269,64 @@ pub unsafe extern "C-unwind" fn gzgets(file: gzFile, buf: *mut c_char, len: c_in
...
+
+    // Change compression parameters for subsequent input.
+    if !state.input.is_null() {
+        // Flush previous input with previous parameters before changing.
+        if state.stream.avail_in != 0 && gz_comp(state, Z_BLOCK).is_err() {
+            return state.err;
+        }
+        // Safety: Because `state` is in write mode and `state.input` is non-null, `state.stream`
+        // was initialized using `deflateInit2` in `gz_init`.
+        unsafe { super::deflateParams(&mut state.stream, level, strategy as c_int) };
```
folkertdev, line 2323: it's really odd that the return value of this function is ignored. Basically, the user is responsible for providing values that are within the allowed ranges, and otherwise this function just silently fails.

Can you add a comment to that effect to the documentation of this function?

folkertdev (approved):

folkertdev:
> @folkertdev this isn't an issue for this PR, but do you have a preferred way to implement C FFI varargs? One of the few gz functions left to implement is:
> 
> ```
> Z_EXTERN int Z_EXPORTVA gzprintf(gzFile file, const char *format, ...);
> ```
> 
> There's a `va_list` in core:ffi that I could use for this, but it's still experimental in the current Rust version.

Hmm yeah that won't work. That feature is experimental, and I don't think it's anywhere near stabilization.

So I think there are 3 options

- add a `"nightly"` feature flag, that can use nightly features. 
- just don't support this function
- do something cursed with `global_asm!`

None of those are good options, really. @bjorn3 do you have thoughts here?

PR #357, opened 2025-05-02
see if using all cores gives us more fuzzer coverage

folkertdev:


folkertdev:
Nope. well good to know

PR #358, opened 2025-05-02
Implement gzseek for reads, as well as gzrewind

brian-pane:


brian-pane:
I'll push a fix for the Windows compilation error (it might be a day or so before I'm able to finish it).

brian-pane:
I only included the read paths for `gzseek` because this PR is kind of big already. I'm planning to add the write support as a separate PR in the future.

PR #359, opened 2025-05-04
Implement gzseek for writes

brian-pane:


libz-rs-sys/src/gz.rs
```diff
@@ -1651,6 +1649,40 @@ unsafe fn gz_write(state: &mut GzState, mut buf: *const c_void, mut len: usize)
...
+    }
+
+    // Compress `len` zeros.
+    let mut first = true;
+    while len != 0 {
+        let n = cmp::min(state.in_size, len);
+        if first {
+            // TODO safety comment
+            let input_slice = unsafe { slice::from_raw_parts_mut(state.input, n) };
+            input_slice.fill(0u8);
```
folkertdev, line 1671: this is almost certainly unsound. Would https://doc.rust-lang.org/std/primitive.pointer.html#method.write_bytes work?

PR #363, opened 2025-05-13
Additional gz test coverage

brian-pane:


brian-pane:
Separate from this PR, the code coverage report shows gaps that I can't figure out. For example, [here](https://app.codecov.io/gh/trifectatechfoundation/zlib-rs/pull/363/blob/libz-rs-sys/src/gz.rs#L584) it shows that `gzclose_r` isn't invoked with a null argument, but that happens [here](https://github.com/trifectatechfoundation/zlib-rs/blob/39838838ec2d49021548f90cec60cc3d8f56b188/test-libz-rs-sys/src/gz.rs#L98) in the test cases.

folkertdev:
You can run cargo llvm-cov locally on just the test-libz-rs-sys I to hlget more accurate coverage I believe.

Also we're at rustweek this week, so I'll probably only get to reviewing this next week

PR #365, opened 2025-05-20
Move gz API to libz-rs-sys-cdylib

brian-pane:


brian-pane:
This currently can compile the gz features with `libz-rs-sys-cdylib && cargo build --features=gz` but the build from the base directory fails due to the C and Rust allocators both being defined.

brian-pane:
With the change I just pushed, the basic build and tests work. If this approach looks safe, I'll troubleshoot the remaining issues with Clippy and fuzz tests.

folkertdev:
let's explore this. I think one of the CI issues is that the cdylib crate should check in its `Cargo.lock`. There also appears to be some bug with the coverage profile in the fuzzers, but that is not triggered by this PR.

@bjorn3 do you have objections to this approach?

brian-pane:
The cdylib's `Cargo.lock` is checked in, and I added a `cargo fetch`, but the `cargo cinstall --offline --release --destdir=/tmp/cargo-cbuild-libz-rs` step still fails because it can't find the `libc` crate.

Some of the other CI failures happen because both `c-allocator` and `rust-allocator` are enabled, but I don't know how the allocator configuration is supposed to work. This [line](https://github.com/trifectatechfoundation/zlib-rs/blob/fdbd002755610848eb79eb3e49e8e3c1f76a4ef1/test-libz-rs-sys/Cargo.toml#L21) in `test-libz-rs-sys/Cargo.toml` seems to turn on both allocators, but I can't see how that would work.

folkertdev:
> This [line](https://github.com/trifectatechfoundation/zlib-rs/blob/fdbd002755610848eb79eb3e49e8e3c1f76a4ef1/test-libz-rs-sys/Cargo.toml#L21) in test-libz-rs-sys/Cargo.toml seems to turn on both allocators, but I can't see how that would work.

It does work for `zlib-rs` (it just means: compile in both versions, rust has precedence but we do make sure it all compiles). It is only in `libz-rs-sys` that we disallow configuring both allocators.

libz-rs-sys-cdylib/Cargo.toml
```diff
@@ -28,9 +28,14 @@ rust-allocator = ["libz-rs-sys/rust-allocator", "libz-rs-sys/std"] # by default,
 custom-prefix = ["libz-rs-sys/custom-prefix"] # use the LIBZ_RS_SYS_PREFIX to prefix all exported symbols
 semver-prefix = ["libz-rs-sys/semver-prefix"] # prefix all symbols in a semver-compatible way
 capi = []
+gz = ["dep:libc"] # gzip support has to be configured explicitly, to avoid a libc dependency in the core zlib-rs
+export-symbols = [] # whether the zlib api symbols are publicly exported
```
bjorn3, line 32: Nothing seems to be using this feature.

folkertdev: yeah, the point of this crate is to expose the symbols, so we don't need that feature flag here.

brian-pane: Do we still need the associated function annotations in `gz.rs`?
```
#[cfg_attr(feature = "export-symbols", export_name = crate::prefix!(gzopen))]
pub unsafe extern "C-unwind" fn gzopen(path: *const c_char, mode: *const c_char) -> gzFile {
```
etc?

brian-pane:
I got the `cargo-c dynamic library` build working, but I'm stuck on the Clippy failures, which are due to both `c-allocator` and `rust-allocator` being defined. I can't tell where in the various `Cargo.toml` the conflicting allocators are getting enabled for `libz-rs-sys`.

libz-rs-sys-cdylib/Cargo.toml
folkertdev: Those can be just `#[export_name = crate::prefix!(gzopen)]` etc. for now

libz-rs-sys-cdylib/src/lib.rs
```diff
@@ -2,5 +2,60 @@ extern crate libz_rs_sys;
 
 pub use libz_rs_sys::*;
 
-#[cfg(not(panic = "abort"))]
-compile_error!("panic=\"abort\" is mandatory because unwinding to C is undefined behavior");
+// TODO somehow make this work with the tests
+// #[cfg(not(panic = "abort"))]
+// compile_error!("panic=\"abort\" is mandatory because unwinding to C is undefined behavior");
```
bjorn3, line 7: Does `#[cfg(not(any(test, panic = "abort"))]` work?

libz-rs-sys-cdylib/src/gz.rs
```diff
@@ -186,7 +189,9 @@ enum Source {
 ///
 /// The caller must ensure that `path` and `mode` point to valid C strings. If the
 /// return value is non-NULL, caller must delete it using only [`gzclose`].
-#[cfg_attr(feature = "export-symbols", export_name = crate::prefix!(gzopen))]
+///
+/// [`gzfree`]: crate::z_stream
+#[export_name = crate::prefix!(gzopen)]
```
bjorn3, line 194: I think this will have to stay mangled during testing to avoid overwriting the C versions of these functions.

bjorn3: The test crate uses both the C and Rust versions of these symbols, right?

folkertdev: Oh, right. I'm now setting the `semver-prefix` feature for `libz-rs-sys-cdylib` in the `test-libz-rs-sys` `Cargo.toml`, which will mangle the names when testing so that they don't clash with the C symbols.

PR #366, opened 2025-05-20
use an autovectorized `slide_hash_chain` on neon

folkertdev:
An engineer at arm confirmed that the slightly different assembly output is equivalent in performance.

This is the assembly we're seeing: https://godbolt.org/z/Kvf7zedeh.


bjorn3 (approved):

folkertdev:
the latest rust release (which bumps llvm) breaks something in the fuzz code coverage. This PR has nothing to do with that, so I'm just going to merge this change.

PR #371, opened 2025-05-27
faster `compare256` aarch64 neon implementation

Dr-Emann:
load 64 bytes at a time, and make a 64-bit bitset of the matches


folkertdev:
Neat, do you have any indication of how much faster this is?

Dr-Emann:
https://gist.github.com/folkertdev/0561efe5f779f5b5cd57d14cc57fbe18

On my mac (m1 pro), using this vs the previous aarch64 implementation goes from ~5 to 2 ns per iteration.

PR #376, opened 2025-05-31
add `gzprintf`

folkertdev:


libz-rs-sys-cdylib/src/gz.rs
```diff
@@ -2571,6 +2571,93 @@ pub unsafe extern "C-unwind" fn gzrewind(file: gzFile) -> c_int {
...
+            format: *const c_char,
+            va: core::ffi::VaList,
+        ) -> i32;
+    }
+
+    let next = state.input.wrapping_add(
+        (unsafe { state.stream.next_in.offset_from(state.input) })
+            .wrapping_add(state.stream.avail_in as isize) as usize,
+    );
+    let state_size = state.in_size / 2;
```
folkertdev, line 2635: This fixes an out-of-bounds read. This makes some sense to me intuitively, but I'm not sure it is correct. We should really document how `in_size` and `out_size` relate to the `size` used by zlib-ng. Can you provide some details there?

folkertdev (commented):
cc @brian-pane, does this implementation look right to you?

libz-rs-sys-cdylib/src/gz.rs
brian-pane: The printf implementation in zlib-ng depends on this:
```
    /* do the printf() into the input buffer, put length in len -- the input
       buffer is double-sized just for this function, so there is guaranteed to
       be state->size bytes available after the current contents */
```

which suggests that zlib-ng's other write functions only ever half-fill the `in` buffer. That's currently not the case for zlib-rs, where the other gz-related write functions currently allow the `input` buffer to fill up completely before flushing it.

folkertdev: I'd prefer to follow `zlib-ng` here. I don't think performance is really a concern.

folkertdev: how hard would it be to instead have `in_size` be the same value as `zlib-ng`? (and allocate twice that number of bytes in read mode, similarly for write)

brian-pane: #378

libz-rs-sys-cdylib/src/gz.rs
```diff
@@ -2595,6 +2595,95 @@ pub unsafe extern "C-unwind" fn gzrewind(file: gzFile) -> c_int {
...
+            n: libc::size_t,
+            format: *const c_char,
+            va: core::ffi::VaList,
+        ) -> i32;
+    }
+
+    // The size of the input buffer is `2 * state.in_size`, just for this function.
+    let in_use = unsafe { state.stream.next_in.offset_from(state.input) } as usize
+        + state.stream.avail_in as usize;
+    let next = state.input.wrapping_add(in_use);
```
brian-pane, line 2658: Is this equivalent to
```
let next = state.stream.next_in.wrapping_add(state.stream.avail_in as usize);
```
?

folkertdev: hmm yeah? this is the original

```c
    if (strm->avail_in == 0)
        strm->next_in = state->in;
    next = (char *)(state->in + (strm->next_in - state->in) + strm->avail_in);
```

libz-rs-sys-cdylib/src/gz.rs
```diff
@@ -2595,6 +2595,95 @@ pub unsafe extern "C-unwind" fn gzrewind(file: gzFile) -> c_int {
...
+
+    // Set the final allocated byte to 0.
+    unsafe { *next.add(state.in_size - 1) = 0 };
```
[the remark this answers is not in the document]
folkertdev: I guess because `zlib-ng` will fall back to `vsprintf` in some cases, which has much weaker safety guarantees.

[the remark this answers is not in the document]
folkertdev: We will not be doing that, so yeah we can get rid of that logic then

libz-rs-sys-cdylib/src/gz.rs
```diff
@@ -2595,6 +2595,111 @@ pub unsafe extern "C-unwind" fn gzrewind(file: gzFile) -> c_int {
...
+            n: libc::size_t,
+            format: *const c_char,
+            va: core::ffi::VaList,
+        ) -> i32;
+    }
+
+    // The size of the input buffer is `2 * state.in_size`, just for this function.
+    // That means we have at least `state.in_size` bytes available.
+    let next = (state.stream.next_in)
+        .wrapping_add(state.stream.avail_in as usize)
```
bjorn3, line 2670: Why not use `add()`? The only difference is that `wrapping_add()` doesn't require the pointer to stay inside the bounds of the allocation, but the `vsnprintf` call requires that anyway.

libz-rs-sys-cdylib/src/gz.rs
```diff
@@ -2595,6 +2595,111 @@ pub unsafe extern "C-unwind" fn gzrewind(file: gzFile) -> c_int {
...
+        .cast_mut();
+
+    // NOTE: zlib-ng writes a NULL byte to the last position of the input buffer. It must do so
+    // because in some cases it falls back to the `vsprintf` function, which contrary to
+    // `vsnprintf` does not guarantee NULL-termination.
+    //
+    // We do not support using `vsprintf`, and therefore don't need to write or check that byte.
+
+    let len = unsafe { vsnprintf(next.cast::<c_char>(), state.in_size, format, va) };
```
bjorn3, line 2679: Can you add a safety comment?

bjorn3 (changes requested):

libz-rs-sys-cdylib/src/gz.rs
```diff
@@ -2595,6 +2595,116 @@ pub unsafe extern "C-unwind" fn gzrewind(file: gzFile) -> c_int {
...
+        .add(state.stream.avail_in as usize)
+        .cast_mut();
+
+    // NOTE: zlib-ng writes a NULL byte to the last position of the input buffer. It must do so
+    // because in some cases it falls back to the `vsprintf` function, which contrary to
+    // `vsnprintf` does not guarantee NULL-termination.
+    //
+    // We do not support using `vsprintf`, and therefore don't need to write or check that byte.
+
+    // This function is not currently exposed by libc, becasue `core::ffi::VaList` is unstable.
```
bjorn3, line 2671: *because

PR #378, opened 2025-05-31
Reserve half of the GzState input buffer space for gzprintf.

brian-pane:
This matches the semantics of zlib-ng: when we allocate a double-sized input buffer in write mode, normal gz write functions are only allowed to write to the front half of the buffer. The remaining space is reserved for gzprintf.

### DIFF
--- a/libz-rs-sys-cdylib/src/gz.rs
+++ b/libz-rs-sys-cdylib/src/gz.rs
@@ -51,7 +51,7 @@ struct GzState {
     source: Source,
     want: usize,     // requested buffer size, default is GZBUFSIZE
     input: *mut u8,  // input buffer (double-sized when writing)
-    in_size: usize,  // size of *input
+    in_size: usize,  // usable size of input buffer (See [`gz_init`] for explanation.)
     output: *mut u8, // output buffer (double-sized when reading)
     out_size: usize, // size of *output
     direct: bool,    // true in pass-through mode, false if processing gzip data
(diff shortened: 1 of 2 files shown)


brian-pane:
That particular line of zlib-ng is generating an offset relative to the `out` buffer. It's in a function that's only used in read mode, which means that the `out` buffer's size is actually `2 * size`. The zlib-rs equivalent uses `out_size`, which is set once to the actual size of the `output` buffer for the mode, instead of the magic `size * 2`. I found it easier to reason about the bounds checks by using `out_size` (which always matches the allocated `output` buf size) instead of `2 * size` in read operations and `size` in write operations. That part is unchanged with this PR.

The same used to apply to `in_size`: it was always the size of the allocated `input` buffer. With this patch, that's no longer the case. Now the semantics are:
* In read mode, `in_size` is the full size of the allocated `input` buffer, which matches what `size` represents in zlib-ng.
* In write mode, `in_size` is half the size of the allocated `input` buffer, which also matches what `size` represents in zlib-ng.

So for read operations, which don't get a double-size `input` buffer, we don't need to double `in_size`. For write operations, we _usually_ don't need to double it, since we're only half-filling the buffer. There are three exception cases where we need to refer to `in_size * 2`:
* When allocating the `input` buffer in write mode, we need to double the requested size. This PR does that by using the new `in_capacity()` to compute the allocation size.
* When deallocating the `input` buffer, we use `in_capacity()` again to match the allocation.
* `gzprintf`, as the one thing allowed to write beyond the halfway point, will need to use `in_capacity()` instead of `in_size` to compute the total available buffer space.

PR #385, opened 2025-06-10
print the zlib.h version used

folkertdev:


folkertdev:
The header files were modified as follows:

- the base is zlib-ng (specifically https://github.com/zlib-ng/zlib-ng/releases/tag/2.2.4, https://github.com/zlib-ng/zlib-ng/commit/860e4cff7917d93f54f5d7f0bc1d0e8b1a3cb988)
- the logic around Z_PREFIX is copied from zlib (so, stock zlib). It looks like zlib-ng uses autoconf magic there, so I went with the simpler approach. (specifically https://github.com/madler/zlib/commit/5a82f71ed1dfc0bec044d9702463dbdf84ea3b71)
- I changed the version to match our `1.3.0.zlib-rs-0.5.1`. (it'll be a bit of a pain to keep that up-to-date, though we don't really need to because the api won't change of course)
- substituted `zlib-rs` for `zlib-ng` in one more spot

I've also now copied the header files to `libz-rs-sys` and made sure with `cargo package --list` that they are included in the `.crate` file. @jeffvanderstoep requested this, it saves them from including an extra crate, and the size of that header file (compressed with zlib-rs, actually) should be minimal.

PR #386, opened 2025-06-12
Export crc32_combine64 and adler32_combine64 for 32bit architectures

jakos-sec:
If `_FILE_OFFSET_BITS` is set to 64 (mostly used on 32bit architectures), it forces the use of 64bit offsets.

The zlib headers used together with libz-rs-sys, redefine `crc32_combine` as `crc32_combine64` (respectively for `adler32_combine`) but they are no longer found in the compiled zlib-rs library. To do so, these functions need to be defined with `z_off64_t` (forced to 64bit).

Since this is my first PR on zlib-rs, I thought I'll open it as is for now, please let me know if I should add some tests or it's not the right direction.

### DIFF
--- a/libz-rs-sys/src/lib.rs
+++ b/libz-rs-sys/src/lib.rs
@@ -99,6 +99,8 @@ pub type z_off_t = c_long;
 #[cfg(target_arch = "wasm32")]
 pub type z_off_t = i64;
 
+pub type z_off64_t = i64;
+
 /// Calculates the [crc32](https://en.wikipedia.org/wiki/Computation_of_cyclic_redundancy_checks#CRC-32_algorithm) checksum
 /// of a sequence of bytes.
 ///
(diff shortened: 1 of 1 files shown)


folkertdev (approved):
Excellent, thanks! 

I did see we're missing some other `*64` functions, so we'll need to add those. Windows is also a bit special apparently, fun.

jakos-sec:
Indeed there are some for the gz* functions, but I didn't find those in `libz-rs-sys/*` yet. I'm happy to add those once they are there, if that is of any help.

folkertdev:
Right, those functions are currently over in https://github.com/trifectatechfoundation/zlib-rs/blob/main/libz-rs-sys-cdylib/src/gz.rs, because they need some extra testing. Eventually they'll move over into the main crate and be re-exported from the `cdylib` crate.

jakos-sec:
That makes sense, thanks for the clarification! Very exciting to see all those "missing" functions being added in rapid speed :)

PR #389, opened 2025-06-17
libz-rs-sys: Build static and shared variants

j-vanderstoep:
Used by the Android project to inform our import tools to generate the build rules necessary for linking this crate from C/C++.

### DIFF
--- a/libz-rs-sys/Cargo.toml
+++ b/libz-rs-sys/Cargo.toml
@@ -22,3 +22,6 @@ semver-prefix = ["export-symbols"] # prefix all symbols in a semver-compatible w
 
 [dependencies]
 zlib-rs = { workspace = true, default-features = false }
+
+[lib]
+crate-type = ["cdylib", "lib", "staticlib"]


folkertdev:
Jeff has previously told me that for android they prefer just depending on `libz-rs-sys` instead of adding an additional dependency on the cdylib crate. 

So far that has not caused issues, but here I do worry that in this case the config makes it seem like you could build a cdylib from the `libz-rs-sys` crate, but the default commands would create an unsound result. As the test failures show, the `libz-rs-sys` crate is not really set up for that either, hence the `cdylib` crate.

bjorn3:
>We always use panic=abort and the system allocator.

It seems like the rust-allocator and std features are being enabled. Or are you overriding this somewhere else? https://cs.android.com/android/kernel/superproject/+/common-android-mainline:external/rust/android-crates-io/crates/libz-rs-sys/Android.bp;l=26 and https://cs.android.com/android/kernel/superproject/+/common-android-mainline:external/rust/android-crates-io/crates/libz-rs-sys/rust_ffi.bp.fragment;l=13;bpv=1;bpt=0

j-vanderstoep:
> It seems like the rust-allocator and std features are being enabled. Or are you overriding this somewhere else? https://cs.android.com/android/kernel/superproject/+/common-android-mainline:external/rust/android-crates-io/crates/libz-rs-sys/Android.bp;l=26 and https://cs.android.com/android/kernel/superproject/+/common-android-mainline:external/rust/android-crates-io/crates/libz-rs-sys/rust_ffi.bp.fragment;l=13;bpv=1;bpt=0

We tell Rust to use libc's allocator by using clang in the same configuration that is used to link C code as our linker. Whichever allocator is selected here should lead to the system allocator - Scudo.

> Do you not need the gz* functions?

I do think that we eventually will want them. Over email, Folkert mentioned "The current batch of `gz*` functions will be moved over to `libz-rs-sys` once we've been able to test them more thoroughly. These functions do a lot more interaction with libc, so there is a bunch of tricky unsafe code that we'd like to validate first." And it sounds like the current implementation requires unstable features. We only support stable compiler versions.

PR #392, opened 2025-06-19
experiment with the `SymBuf` representation

folkertdev:
@brianpane you might enjoy playing around with this.

I get some actual speedup for compression level 3, it appears to be pretty neutral for the others. In `zlib-ng`, they actually use 2 arrays, which I did try but only gave instruction count improvements, not cycles/wall time.

I think the padding field is important because it makes initialization as just all zeros possible, which is efficient for the `fill` operations.

but well there is still the ordering of the fields, maybe `#[repr(packed)]` is useful, and maybe there are other ideas still.

### DIFF
--- a/zlib-rs/src/deflate.rs
+++ b/zlib-rs/src/deflate.rs
@@ -1,6 +1,8 @@
 #![warn(unsafe_op_in_unsafe_fn)]
 use core::{ffi::CStr, marker::PhantomData, mem::MaybeUninit, ops::ControlFlow};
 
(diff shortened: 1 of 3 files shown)


folkertdev:
the https://github.com/trifectatechfoundation/zlib-rs/tree/sym-buf-experiment branch has some further experiments. I can't really pin it down and get conclusive improvements.

folkertdev:
I don't think anything is going to come of this at this point.

PR #396, opened 2025-07-17
Strengthen the lint directives related to unsafe code

brian-pane:
* Make deny(unsafe_op_in_unsafe_fn) the default
* Add forbid(unsafe_code) to hash_calc.rs now that it no longer contains unsafe code
* Document a few additional safety preconditions

### DIFF
--- a/zlib-rs/src/allocate.rs
+++ b/zlib-rs/src/allocate.rs
@@ -1,4 +1,5 @@
 #![allow(unpredictable_function_pointer_comparisons)]
+#![allow(unsafe_op_in_unsafe_fn)]
 
 #[cfg(unix)]
 use core::ffi::c_int;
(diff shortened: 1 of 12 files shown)


bjorn3:
`#[deny(unsafe_op_in_unsafe_fn)]` should be fine.

zlib-rs/src/allocate.rs
```diff
@@ -442,6 +445,14 @@ mod tests {
...
     }
 
+    #[test]
+    fn align_size() {
+        // posix_memalign requires that the alignment be both a power of two and
+        // a multiple of sizeof(void*).
+        assert_eq!(ALIGN.count_ones(), 1);
+        assert_eq!(ALIGN as usize % mem::size_of::<*mut c_void>(), 0);
+    }
+
```
folkertdev, line 455: you can write this as 

    const _: () = assert!(ALIGN.count_ones() == 1);
    const _: () = assert!(ALIGN as usize % mem::size_of::<*mut c_void>() == 0);

then they are evaluated at compile time

PR #397, opened 2025-07-17
Declare BitReader::refill unsafe and document safety prerequisites

brian-pane:


zlib-rs/src/inflate.rs
```diff
@@ -1862,15 +1862,31 @@ fn inflate_fast_help_impl<const FEATURES: usize>(state: &mut State, _start: usiz
     let mut bad = None;
 
     if bit_reader.bits_in_buffer() < 10 {
-        bit_reader.refill();
+        if bit_reader.bytes_remaining() >= 8 {
```
brian-pane, line 1865: I added this check to make sure the safety precondition was met. But if there's any way to know for sure that `buffer.bytes_remaining()` is guaranteed to be >= 8 when this function is called, I can document that with a comment in place of the if-statement here.

brian-pane:
By following the calling functions, I was able to document why the very first call to `refill` in `inflate_fast_help_impl` is safe: all the calling locations ensure that the `bit_reader` has at least 15 bytes left, when we only need 8 for the `refill`.

zlib-rs/src/inflate.rs
```diff
@@ -1817,30 +1821,49 @@ impl State<'_> {
     }
 }
 
-fn inflate_fast_help(state: &mut State, start: usize) {
+// # Safety
+//
+// `state.bit_reader` must have at least 8 bytes available to read, as
+// indicated by `state.bit_reader.bytes_remaining() >= 8`
```
bjorn3, line 1827: Please use doc comments.

folkertdev:
Neat, great work! This looks good but I want to look at it with fresher eyes, should be able to do that tomorrow.

PR #403, opened 2025-08-05
Add safety comments to Writer implementation

brian-pane:


zlib-rs/src/inflate/writer.rs
```diff
@@ -272,27 +285,47 @@ impl<'a> Writer<'a> {
     /// # Safety
     ///
     /// `src` must be safe to perform unaligned reads in `core::mem::size_of::<C>()` chunks until
```
brian-pane, line 287: Should this be `core::mem::size_of::<N>` rather than `<C>`?

PR #405, opened 2025-08-26
clippy

folkertdev:
kind of silly, but :shrug:

Ideally we'd use an if-let chain here, but our msrv does not allow that of course.


bjorn3:
I would just disable the clippy lint here or remove the `&& window_offset.is_some()` and move the comment into the `if let` body.

PR #408, opened 2025-09-01
work around avx512vl miscompilation

### LINKED ISSUE #407: Bug in crc32 calculations
Hopefully this is not a case of [PEBKAC](https://en.wikipedia.org/wiki/User_error), but I believe that crc32 calculations are not working correctly. I initially found this when running against CPython's test suite trying to build CPython using zlib-rs. The failing test case was https://github.com/python/cpython/blob/552cf86b83e8aafebcfe9abbc77101bdbfda14c8/Lib/test/test_zlib.py#L169-L174, which I've minimized to the following reproducer, which passes when linked with zlib, but not libz-rs-sys-cdylib:

```C
#include <stdio.h>
#include <string.h>
#include "zlib.h"

int main()
{
    unsigned char a[32] = {0x88, 0x64, 0x15, 0xce, 0x5e, 0x3b, 0x8d, 0x35,
                           0xdb, 0xd2, 0xb5, 0xfa, 0x8e, 0xa7, 0x73, 0x10,
                           0x66, 0x83, 0x1b, 0xd1, 0xde, 0x0f, 0x25, 0x86,
                           0xeb, 0xe5, 0x42, 0x44, 0xad, 0x62, 0xff, 0x11};
    uInt chk_a = crc32(0, a, 32);
    unsigned char b[64] = {0x31, 0xb8, 0xce, 0x94, 0x4d, 0x2b, 0xb9, 0x7e,
                           0xd5, 0x81, 0x7f, 0xc2, 0x40, 0xbf, 0x3d, 0xa5,
                           0x25, 0xa5, 0xf9, 0xdf, 0x53, 0x68, 0xc4, 0xf6,
                           0xbe, 0x06, 0x7d, 0xf3, 0xc7, 0xdc, 0x5b, 0x84,
                           0xce, 0xd2, 0xb2, 0xeb, 0x87, 0x62, 0x60, 0xe3,
                           0x10, 0x05, 0x64, 0x59, 0x15, 0xc4, 0x2d, 0x78,
                           0xc8, 0xf3, 0x14, 0x38, 0x87, 0x39, 0xb3, 0x58,
                           0xb5, 0x95, 0x07, 0x25, 0xd9, 0xc1, 0xac, 0x04};
    uInt chk_b = crc32(0, b, 64);
    unsigned char buff[96];
    memcpy(buff, a, 32);
    memcpy(buff + 32, b, 64);
    uInt chk = crc32(0, buff, 96);
    uInt chk_combine = crc32_combine(chk_a, chk_b, 64);
    printf("chk (%u) = chk_combine (%u)? %s\n", chk, chk_combine, chk == chk_combine ? "True" : "False");
    return (0);
}
```

Example output:
```
$ clang-20 -o test test.c -l z_rs && ./test
chk (2686788294) = chk_combine (2546675245)? False
$ clang-20 -o test test.c -lz && ./test
chk (2546675245) = chk_combine (2546675245)? True
```

folkertdev:
fixes #407 

Extracting this inner loop and making sure it does not inline seems to fix the issues we're seeing.

The performance impact should be minimal: the loop is still inlined on non-x86_64 targets, and even by default when compiling for x86_64. Unfortunately, we need to use avx2 as a proxy for this avx512vl target feature, because it has only been stable since 1.89.0, while the earliest affected version is 1.87.0.

We suspect this to be an LLVM problem, though I've not been able to pinpoint the exact LLVM issue.

cc @emmatyping

### DIFF
--- a/zlib-rs/src/crc32/braid.rs
+++ b/zlib-rs/src/crc32/braid.rs
@@ -103,10 +103,7 @@ pub fn crc32_braid<const N: usize>(start: u32, data: &[u8]) -> u32 {
 
         crcs.fill(0);
         for j in 0..W {
-            for k in 0..N {
-                crcs[k] ^= Crc32BraidTable::<N>::TABLE[j][buffer[k] & 0xff];
-                buffer[k] >>= 8;
-            }
+            braid_core(&mut crcs, &mut buffer, j);
         }
     }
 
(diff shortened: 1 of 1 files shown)


folkertdev:
@emmatyping does the python test suite work with this fix? If so we'll make a release.

PR #409, opened 2025-09-02
export `get_crc_table`

folkertdev:
this is an undocumented function, but one that is used in practice. It's easy enough to support


folkertdev:
Well as far as I can tell it's officially undocumented https://github.com/zlib-ng/zlib-ng/blob/83a84faf7a67cb9926947d6b618f141c8020e8e3/zlib.h.in#L1838-L1846, but as that page shows, it is used in practice.

PR #410, opened 2025-09-04
remove `fn min_max` in `inftrees.rs`

folkertdev:
turns out this function was sort of fully unrolled, emitting ~200 instructions ?!


folkertdev:
subsumed by https://github.com/trifectatechfoundation/zlib-rs/pull/411

PR #425, opened 2025-11-16
add avx512 adler32 implementation

folkertdev:
This will need tomorrow's nightly for miri to support all avx512 instructions we use here


zlib-rs/src/adler32/avx512.rs
```diff
@@ -0,0 +1,217 @@
...
+    16, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1,
+]);
+
+const ZERO: __m512i = __m512i_literal([0u8; 64]);
+
+#[target_feature(enable = "avx512f")]
+#[target_feature(enable = "avx512bw")]
+pub fn adler32_avx512(adler: u32, src: &[u8]) -> u32 {
+    assert!(cfg!(target_feature = "avx512f"));
+    assert!(cfg!(target_feature = "avx512bw"));
```
bjorn3: Calling `adler32_avx512` is already UB if avx512 is not available due to the `#[target_feature(enable = "...")]`. For avx2 you don't have a `#[target_feature(enable = "...")]`.

PR #426, opened 2025-11-18
Inflate back

folkertdev:
So, this is unfortunately a lot of code for 3 functions that I've never seen anyone use. However, it's one fewer argument against zlib-rs, so, fine.

I've fuzzed this quite extensively, so I believe we should be good on that front. The main concern is that we'd mess with the non-back implementation, but in the end i had to copy most interesting functions because `inflateBack` just has different assumptions, so I think this is relatively risk-free.

So mostly I just want to implement this and then forget about it.

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -87,3 +87,9 @@ name = "checksum"
 path = "fuzz_targets/checksum.rs"
 test = false
 doc = false
+
+[[bin]]
+name = "infback"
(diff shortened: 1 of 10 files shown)


bjorn3 (approved):
I mostly skipped reviewing the testing and fuzzing code.

folkertdev:
Yeah, fair. We get decent coverage, so it should be fine.

PR #430, opened 2025-11-29
Inflate: use a single allocation

folkertdev:


zlib-rs/src/inflate.rs
```diff
@@ -2192,31 +2235,38 @@ pub fn init(stream: &mut z_stream, config: InflateConfig) -> ReturnCode {
...
     // SAFETY: we've correctly initialized the stream to be an InflateStream
-    let ret = if let Some(stream) = unsafe { InflateStream::from_stream_mut(stream) } {
-        reset_with_config(stream, config)
+    if let Some(stream) = unsafe { InflateStream::from_stream_mut(stream) } {
+        stream.state.allocation_start = allocation_start.as_ptr();
+        stream.state.total_allocation_size = allocs.total_size;
+        let ret = reset_with_config(stream, config);
+
+        if ret != ReturnCode::Ok {
+            end(stream);
```
spectral54: I may be interpreting the code incorrectly, but I think this has the possibility for a memory leak if `from_stream_mut(stream)` returns None, as we won't call `end()`.

Previously if `from_stream_mut()` returned None, then we set `ret = ReturnCode::StreamError`, and *then* checked if it was Ok (it wasn't) and called `deallocate`. Now, the only path to calling `deallocate` requires that `from_stream_mut` has returned Some.

We discovered this when importing this new version into our codebase. Complicating factors: we use a custom malloc, so I'm not 100% sure this isn't an issue with that, or with our wrappers around this (to use zlib-rs from C++ code), or it might be a false positive in our leak detection stuff, as I wasn't able to reproduce the leak when I ran it with AddressSanitizer (which has its own leak detection stuff, I think?)

folkertdev: Given the definition

```rust
    pub unsafe fn from_stream_mut(strm: *mut z_stream) -> Option<&'a mut Self> {
        {
            // Safety: ptr points to a valid value of type z_stream (if non-null)
            let stream = unsafe { strm.as_ref() }?;

            if stream.zalloc.is_none() || stream.zfree.is_none() {
                return None;
            }

            if stream.state.is_null() {
                return None;
            }
        }

        // Safety: InflateStream has an equivalent layout as z_stream
        unsafe { strm.cast::<InflateStream>().as_mut() }
    }
```

I don't think we can reasonably free anything if that function returns `None`. Either

- the stream pointer is not valid for the type (e.g. misaligned)
- no consistent allocator has been provided, so best not to call some potentially garbage function pointer
- there is no state (i.e. we don't have any address of an allocation that we can free)

So if you're seeing a real problem, I don't think it's here. Or if it is, then your stream/state are already in some invalid configuration.

spectral54: Hmm, yeah this does seem to be an incorrect root cause analysis on my part. I was too focused on init() because the stack trace we saw was (roughly, I removed a bunch of stuff):

```
Leak of 80384 bytes in 1 objects allocated from:
	@ heap-checker.cc
	@ tcmalloc/malloc_hook.cc
	@ tcmalloc/tcmalloc.cc __posix_memalign
	@ __rustc::__rdl_alloc
	@ <zlib_rs::allocate::Allocator>::allocate_layout
	@ zlib_rs::inflate::init
	@ rs_inflateInit2_
	@ riegeli/zlib/zlib_reader.cc
	@ <other stuff>
```

But that's just where it was allocated from, not where it was leaked from. Let me do some more work on my side to try to figure out how precisely where we're holding this incorrectly, I think that's probably the more likely scenario now. If I have proof (in the form of a reproduction test that I can contribute to zlib-rs) that this is some issue in the zlib-rs code, I'll file an issue / send a PR. Sorry for the noise!

PR #437, opened 2025-12-03
Set Z_HAVE_UNISTD_H only on non-Windows platforms

jakos-sec:
`unisdt.h` does not exist on Windows and therefore `Z_HAVE_UNISTD_H` should not be set.

When trying to use zlib-rs on Windows I received the following:
```
fatal error: 'unistd.h' file not found
  316 | #  include <unistd.h>         /* for SEEK_*, off_t, and _LFS64_LARGEFILE */
      |            ^~~~~~~~~~
1 error generated.
```

### DIFF
--- a/libz-rs-sys/include/zconf.h
+++ b/libz-rs-sys/include/zconf.h
@@ -291,7 +291,7 @@ typedef void       *voidp;
 
 typedef unsigned int z_crc_t;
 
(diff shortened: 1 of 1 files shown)


jakos-sec:
I ~~ think ~~ so. I haven't used the libz-rs-sys-cdylib but I assume it also uses the same headers (zlib.h/zconf.h) as libz-rs-sys?

PR #450, opened 2026-01-01
Replace run-time asserts with compile time in deflate tally functions

brian-pane:


folkertdev:
> Hyperfine reports a 1% improvement in elapsed time in my testing on Mac M2 at deflate level 1, and no change at higher levels.

I can't really measure anything on linux, but I think the changes look good overall. I'd assume LLVM would be able to take care of this, but maybe it's too many steps.

> Unfortunately I don't have good tooling to measure CPU cycles on macO

Us neither. We've played with the idea of setting up asahi linux on a mac mini so we have linux (and hence, `perf`) on the apple hardware, but haven't gotten to it.

PR #453, opened 2026-01-07
fuzzers: Use bytes instead of string inputs

Mrmaxmeier:
Constraining these to valid UTF-8 doesn't seem useful here


folkertdev (approved):
Nice, thanks!

Issue #459, opened 2026-01-12
Deflate output not deterministic when the compressor is reused with deflateReset


QrczakMK:
Deflate output may depend on previous contents of a `Window`.

Scenario 1: compress data A.

Scenario 2: compress data B, `deflateReset`, compress data A.

Sample data A ends with `"\201\201\201\201\201\000\201\201\260\201\201\201\201\201\201\201\201\201"` (C++ literal syntax; I can include the whole data here if needed, but reproduction program would require too much work to isolate).

The difference in compressed outputs for compression level 6 (levels 2..6 are affected, with slightly different outputs):

```
 ...
 inflate:         length 5
 inflate:         distance 9
 inflate:         length 4
-inflate:         distance 5
+inflate:         distance 13
 inflate:         end of block
 inflate:   check matches trailer
```

Both outputs are correct, but the output should be deterministic.

* * *

Scenarios 1 and 2 start diverging here: https://github.com/trifectatechfoundation/zlib-rs/blob/4398c8ab4248c58d80cf6dbb5115382ade32ad2d/zlib-rs/src/deflate/longest_match.rs#L196-L200

In scenario 1, `scan_val == 0x0000000081818181`, where the 0x81's are real bytes, and the 0x00's are - now I am guessing - garbage bytes from internal state outside the copy of the real data. They are not supposed to affect the output, because they will be verified later.

Trying a match at `cur_match == 111` (distance 13), where `match_val == 0x8181818181818181`. This yields `len == 4`.

Looking for a better match. At `cur_match == 103` (distance 5), `match_val == 0xb081810081818181`. This yields `len == 5`, because the 0x00 in the input data happens to match the meaningless 0x00 after the data! The match looks better. It is checked to exceed input data though: https://github.com/trifectatechfoundation/zlib-rs/blob/4398c8ab4248c58d80cf6dbb5115382ade32ad2d/zlib-rs/src/deflate/longest_match.rs#L257-L260

The search stops, because a yet better match is impossible. The truncated match is emitted, with length 4 at distance 5.

In scenario 2, at this point `scan_val == 0x9898989881818181`. This also contains 4 garbage bytes, but this time they come from an unrelated compression session.

The first match is as before.

Looking for a better match. This time the 0x00 does not match, and `len == 4` again. The match is not better, so the best match so far is still `cur_match == 111` (distance 13).

[...]

PR #461, opened 2026-01-15
add test for deterministic deflate output

folkertdev:
fixes https://github.com/trifectatechfoundation/zlib-rs/issues/459

Compressing an input after a `reset` should give the same output as if it was done with a fresh compression state. There was a bug where some of the state from before the reset could influence what match was picked for later input. The output was still valid, but it is inconsistent.

Thank you @QrczakMK for the detective work here!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ opt-level = 1 # required for the tail calls in inflate to optimize
 
 [workspace.dependencies]
 libloading = "0.8.1"
-libz-sys = { version = "1.1.23", default-features = false, features = ["zlib-ng"] } # use libz-ng in libz compat mode
+libz-sys = { version = "1.1.24", default-features = false, features = ["zlib-ng"] } # use libz-ng in libz compat mode
 arbitrary = { version = "1.0" }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -36,6 +36,7 @@ libloading = "0.8.1"
 libz-sys = { version = "1.1.23", default-features = false, features = ["zlib-ng"] } # use libz-ng in libz compat mode
 arbitrary = { version = "1.0" }
 quickcheck = { version = "1.0.3", default-features = false, features = [] }
+getrandom = "0.2.17" # work around https://github.com/rust-random/getrandom/issues/775
```
bjorn3: I did say this is the proper fix. Something is broken with the older version, so you effectively have a dependency on the newer version already, this is just declaring said dependency.

```suggestion
getrandom = "0.2.17" # Earlier getrandom versions contain UB: https://github.com/rust-random/getrandom/issues/775
```

Does specifying this in `workspace.dependencies` have any effect though? Aren't `workspace.dependencies` and `workspace.package` just inert templates that an actual package can use with `foo.workspace = true`? Maybe move it to the `dev-dependencies` section of `test-libz-rs-sys/Cargo.toml` ?

PR #479, opened 2026-02-26
Run all the gz tests when building on macOS

brian-pane:


brian-pane:
Oh, this doesn't work with WASM or with Miri on Windows. I'll update the PR to just enable building on macOS in addition to Linux.

PR #501, opened 2026-04-16
add uninit versions of stable compress and decompress

folkertdev:
For use in flate2 (https://github.com/rust-lang/flate2-rs/issues/544)

### DIFF
--- a/zlib-rs/src/stable.rs
+++ b/zlib-rs/src/stable.rs
@@ -1,4 +1,4 @@
-use core::ffi::c_uint;
+use core::{ffi::c_uint, mem::MaybeUninit};
 
 use crate::deflate::DeflateConfig;
 use crate::inflate::InflateConfig;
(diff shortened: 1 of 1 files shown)


zlib-rs/src/stable.rs
```diff
@@ -126,12 +126,28 @@ impl Inflate {
...
+    /// with `flush` defining some details about this.
     pub fn decompress(
         &mut self,
         input: &[u8],
         output: &mut [u8],
         flush: InflateFlush,
+    ) -> Result<Status, InflateError> {
+        self.decompress_uninit(
+            input,
+            unsafe { &mut *(output as *mut _ as *mut [MaybeUninit<u8>]) },
```
bjorn3, line 139: Would have been nice if `MaybeUninit::as_bytes_mut` was stable.

bjorn3 (approved):

zlib-rs/src/stable.rs
folkertdev: I don't think that helps by itself

```rust
// core::mem

impl<T> MaybeUninit<T> {
    pub fn as_bytes(&self) -> &[MaybeUninit<u8>];
    pub fn as_bytes_mut(&mut self) -> &mut [MaybeUninit<u8>];
}

impl<T> [MaybeUninit<T>] {
    pub const fn as_bytes(&self) -> &[MaybeUninit<u8>];
    pub const fn as_bytes_mut(&mut self) -> &mut [MaybeUninit<u8>];
}
```

The thing here is that we are turning a `&mut T` into a `&mut MaybeUninit<T>` which you need to be careful not to write uninitialized data into. That conversion can never be safe I think?

bjorn3: Right

PR #512, opened 2026-05-25
fix and test `aarch64_be`

folkertdev:
So this is obviously a tier 3 target that we do not need to support, however:

- we promise to be portable, and hence should try hard to in fact be portable (e.g. for firefox and who knows where we end up)
- as a co-maintainer of stdarch I both would like to notice issues early and can fix them myself when they come up

### DIFF
--- a/zlib-rs/src/adler32/neon.rs
+++ b/zlib-rs/src/adler32/neon.rs
@@ -4,7 +4,7 @@
 //! extension.
 use core::arch::aarch64::{
     uint16x8_t, uint16x8x2_t, uint16x8x4_t, uint8x16_t, vaddq_u32, vaddw_high_u8, vaddw_u8,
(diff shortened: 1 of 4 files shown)


zlib-rs/src/adler32/neon.rs
```diff
@@ -116,9 +116,26 @@ unsafe fn accum32(s: (u32, u32), buf: &[uint8x16_t]) -> (u32, u32) {
 
     let mut it = buf.chunks_exact(4);
 
+    #[target_feature(enable = "neon")]
+    fn swap_64bit_lane_when_be(v: uint8x16_t) -> uint8x16_t {
```
bjorn3:
```suggestion
    #[inline]
    fn swap_64bit_lane_when_be(v: uint8x16_t) -> uint8x16_t {
```

bjorn3 (approved):

PR #515, opened 2026-05-29
crc32: add riscv64 implementation

inahga:
@folkertdev is there any interest in porting zlib-ng's riscv64 implementations into zlib-rs? I recently found myself with some riscv64 hardware, so happy to do so.

Checking if this is something you're willing to accept/maintain before I go further.

### DIFF
--- a/zlib-rs/src/cpu_features.rs
+++ b/zlib-rs/src/cpu_features.rs
@@ -110,9 +110,11 @@ pub fn is_enabled_simd128() -> bool {
 }
 
 #[inline(always)]
-pub fn is_enabled_zbc() -> bool {
+pub fn is_enabled_zbkc() -> bool {
+    // FIXME: std::arch::is_riscv64_feature_detected is stabilized in 1.78. Switch to runtime
+    // feature detection once MSRV is bumped. Until then, zbkc support is compile-time only.
     #[cfg(target_arch = "riscv64")]
-    return cfg!(target_feature = "zbc");
+    return cfg!(target_feature = "zbkc");
 
     false
 }
--- a/zlib-rs/src/cpu_features.rs
+++ b/zlib-rs/src/cpu_features.rs
@@ -108,3 +108,11 @@ pub fn is_enabled_simd128() -> bool {
 
     false
 }
+
+#[inline(always)]
+pub fn is_enabled_zbc() -> bool {
+    #[cfg(target_arch = "riscv64")]
+    return cfg!(target_feature = "zbc");
+
+    false
+}
--- a/zlib-rs/src/crc32/zbc.rs
+++ b/zlib-rs/src/crc32/zbc.rs
@@ -1,0 +1,107 @@
+//! crc32 implementation using the riscv64 zbc ISA extension. Derived from
+//! zlib-ng's implementation, see
+//! https://github.com/zlib-ng/zlib-ng/blob/da22434b657578c41af1bdf06b27304e4aceb00f/arch/riscv/crc32_zbc.c
+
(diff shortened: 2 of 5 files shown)


zlib-rs/src/crc32/zbc.rs
```diff
@@ -0,0 +1,107 @@
...
+    (ret >> 32) as u32
+}
+
+/// Inline assembly for required instructions, since the intrinsics are nightly
+/// compiler only.
+mod asm {
+    // Returns the lower half of carryless multiplication of rs1 and rs2.
+    // See https://riscv.github.io/riscv-isa-manual/snapshot/spec/#insns-clmul
+    #[inline(always)]
+    pub fn clmul(rs1: u64, rs2: u64) -> u64 {
```
folkertdev: I think this function still needs the appropriate target feature declared?

zlib-rs/src/cpu_features.rs
```diff
@@ -108,3 +108,11 @@ pub fn is_enabled_simd128() -> bool {
 
     false
 }
+
+#[inline(always)]
+pub fn is_enabled_zbc() -> bool {
+    #[cfg(target_arch = "riscv64")]
+    return cfg!(target_feature = "zbc");
```
folkertdev: Can you leave an note about runtime feature detection here? 

https://doc.rust-lang.org/std/arch/macro.is_riscv_feature_detected.html

is stable since 1.78 while our current msrv is 1.75

zlib-rs/src/crc32/zbkc.rs
folkertdev, line 1: Not needed for this PR, but I suspect an approach like in https://github.com/trifectatechfoundation/zlib-rs/pull/514 could work here too? Though aarch64 has `64 x 64 -> 128` bits in one instruction, while here it takes 2 so perhaps it's not actually advantageous.

zlib-rs/src/crc32/zbc.rs
```diff
@@ -0,0 +1,107 @@
...
+    // Barrett reduction step
+    let mut barrett = clmul(reduced_low & MASK32, CONSTANT_RU) & MASK32;
+    barrett = clmul(barrett, CRCPOLY_TRUE_LE_FULL);
+    let ret = barrett ^ reduced_low;
+
+    (ret >> 32) as u32
+}
+
+/// Inline assembly for required instructions, since the intrinsics are nightly
+/// compiler only.
```
folkertdev:
```suggestion
/// Inline assembly for required instructions, since the stdarch intrinsics are currently unstable.
```

zlib-rs/src/crc32/zbc.rs
folkertdev: so, `#[target_feature(enable = "zbkc")]`

folkertdev (commented):
Totally, riscv support just hasn't come up yet. 

Thanks for adding the testing, that is kind of the main requirement I have.

Did you check that the bounds checks actually get eliminated? it looks like they should but that might need some slight tweaks.

Finally, perhaps you can add the clmul instructions to miri? You can combine ideas from

- https://github.com/rust-lang/miri/pull/5070
- https://github.com/rust-lang/miri/pull/5062

For the code changes. We usually like to validate the added test cases on real hardware, which you can actually do.

inahga:
Bad news, the two systems I thought had zbkc, actually don't! And worse yet, they also don't do zvv for compare256/slidehash/adler32. Not insurmountable, more modern dev boards here seem to be at reasonable cost. I'll come back to this after I source one and after my vacation.

PR #525, opened 2026-06-09
Add `clone` to `Deflate` and `Inflate`

### LINKED ISSUE #524: [Feature Request] `inflateCopy` and `deflateCopy` equivalents in the Rust API
I'm not sure if there's a deliberate design decision behind this, but it feels off that the C API `libz-rs-sys` offers more functionality than the pure Rust API given that this library is implemented in Rust. Implementing the `Clone` trait on `Deflate` and `Inflate` seems like an idiomatic fit for exposing the same functionality as provided by `inflateCopy` and `deflateCopy` in the Rust API.

<details>

<summary>Extra context</summary>

There's currently a spike going on trying to get Rust integrated into CPython. As a first goal Python's `zlib` module was chosen to be ported from C to Rust. While using the APIs in `libz-rs-sys` would be an option, it somewhat defeats the purpose by loosing much of the memory- and type-safety benefits Rust gives.

Cf. https://github.com/Rust-for-CPython/zlib-py/issues/4

</details>

robsdedude:
Closes: https://github.com/trifectatechfoundation/zlib-rs/issues/524


zlib-rs/src/stable.rs
```diff
@@ -261,22 +264,16 @@ impl Inflate {
...
             Err(InflateCloneError::MemError) => panic!("allocation failed"),
             Err(InflateCloneError::StreamError) => panic!("Inflate in inconsistent state"),
-            Err(other) => unreachable!("Inflate::try_clone does not return {other:?}"),
         }
     }
-
-    // TODO: optimization potential by implementing clone_from
-    //       Current hurdle is that going from `&mut self.inner` to
-    //       `&mut MaybeUninit<InflateStream<'a>>` in a cheap way requires
-    //       pointer massaging.
```
robsdedude: I realized that there can't be a safe implementation of `try_clone_from` because if `copy` fails with `ReturnCode::MemError`, `core::ptr::copy_nonoverlapping(source, dest)` will already have been called leaving `self` with the same pointers as `dest`/`other`. This effectively turns a shared ref `other: &Self` into a mutable ref `&mut self`.

A possible solution is to make `inflate::copy` and `deflate::copy` "atomic" by attempting the allocation first and only on success starting to actually change `dest`. I'm not sure what the guarantees on the C API are, btw. Are users expected to assume that a failed `copy` leaves `dest` in a "better don't use this" state?

zlib-rs/src/deflate.rs
```diff
@@ -601,7 +601,7 @@ pub fn prime(stream: &mut DeflateStream, mut bits: i32, value: i32) -> ReturnCod
 
 pub fn copy<'a>(
     dest: &mut MaybeUninit<DeflateStream<'a>>,
-    source: &mut DeflateStream<'a>,
```
robsdedude, line 604: It's sufficient if one ref is mutable for the safety comment below around the mem copy. A mutable reference means no other reference to that piece of memory can exist (in safe rust). Further does this method not actually mutate `source`.

zlib-rs/src/deflate.rs
```diff
@@ -601,7 +601,7 @@ pub fn prime(stream: &mut DeflateStream, mut bits: i32, value: i32) -> ReturnCod
 
 pub fn copy<'a>(
```
robsdedude, line 602: Open questions:
 * `inflate::copy` is unsafe, while this one isn't. What's the difference?
 * `inflate::copy`, while marked unsafe, does not document which invariants the caller needs to uphold. I'd appreciate an explanation as I'm unsure.

robsdedude:
@folkertdev If you have some time (no rush), I'd appreciate if you could give this a first look. Especially the open question I've left in one of the comments.

zlib-rs/src/deflate.rs
folkertdev: `inflate_copy` can be marked safe now. Can you also update the comment on the `copy_nonoverlapping` there (we now have a mutable and an immutable reference, but of course those still cannot alias).

zlib-rs/src/stable.rs
```diff
@@ -212,6 +238,42 @@ impl Inflate {
             other => unreachable!("set_dictionary does not return {other:?}"),
         }
     }
+
+    /// Clone the [`Inflate`] state.
+    ///
+    /// This is conceptually like [`Clone::clone`](`core::clone::Clone::clone`),
+    /// except it's fallible.
+    pub fn try_clone(&self) -> Result<Self, InflateCloneError> {
```
folkertdev, line 246: Do you need this? Because if not I'd rather keep it private (or leave it out entirely) to reduce the amount of API surface.

zlib-rs/src/stable.rs
```diff
@@ -66,6 +66,32 @@ impl InflateError {
     }
 }
 
+/// Errors that can occur when cloning [`Inflate`]
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+#[repr(i32)]
+pub enum InflateCloneError {
+    /// The [`Inflate`] is in an inconsistent state, most likely
+    /// due to an invalid configuration parameter.
+    StreamError = -2,
```
folkertdev, line 75: is this reachable with just the stable API?

folkertdev (commented):
Thanks for looking into this!

This'll need some tests in `test-libz-rs-sys/src/stable.rs`

PR #529, opened 2026-06-27
Improve the LoongArch CRC32 implementation based on the AArch64 one

Gelbpunkt:
This ports the stock zlib AArch64 CRC32 implementation over to LoongArch, but accounts for the throughput of 2 rather than 3.

Gains from the sse benchmark show around 10% wall time gain on 1GB input, which is in line with the 15% gain I saw on AArch64. This apparently scales linearly with the instruction throughput.


folkertdev (approved):
Looks great, and our miri testrun tests this implementation. Thanks!

PR #531, opened 2026-06-27
crc32/loongarch: CRC intrinsics are stabilized in 1.98

Gelbpunkt:
We therefore don't need to set feature(stdarch_loongarch) for Miri anymore, since Miri always runs on nightly where it's now stable.


Gelbpunkt:
This depends on stdarch being merged back into rust

folkertdev:
Now that we use more of the loongarch api surface, this PR is no longer needed.

PR #532, opened 2026-06-27
Add LSX-autovectorized slide hash

Gelbpunkt:
See https://godbolt.org/z/jhrGxPn87. In practice, both loongarch64-unknown-linux-gnu and loongarch64-unknown-linux-musl include LSX in their target baseline (though it isn't required by the ISA), so the feature detection is usually optimized away.


Gelbpunkt:
Hmmm, I'd love to run CI with this feature enabled. Not sure how to best go about that

PR #533, opened 2026-06-28
Add LSX-optimized adler32 implementation

Gelbpunkt:
There's for sure a lot left to optimize here, this is a mostly faithful port of the NEON implementation at the moment. Should serve as a good baseline for future improvements.

### DIFF
--- a/zlib-rs/src/adler32.rs
+++ b/zlib-rs/src/adler32.rs
@@ -9,6 +9,8 @@ mod avx512;
 #[cfg(target_arch = "x86_64")]
 mod avx512_vnni;
 mod generic;
+#[cfg(all(target_arch = "loongarch64", feature = "lsx"))]
+mod lsx;
 #[cfg(target_arch = "aarch64")]
 mod neon;
 #[cfg(any(target_arch = "wasm32", target_arch = "wasm64"))]
@@ -34,6 +36,11 @@ pub fn adler32(start_checksum: u32, data: &[u8]) -> u32 {
     #[cfg(any(target_arch = "wasm32", target_arch = "wasm64"))]
     if crate::cpu_features::is_enabled_simd128() {
         return self::wasm::adler32_wasm(start_checksum, data);
+    }
+
+    #[cfg(all(target_arch = "loongarch64", feature = "lsx"))]
+    if crate::cpu_features::is_enabled_lsx() {
+        return self::lsx::adler32_lsx(start_checksum, data);
     }
 
     generic::adler32_rust(start_checksum, data)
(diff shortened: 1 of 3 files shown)


Gelbpunkt:
Awwww we can't run this in Miri quite yet

Gelbpunkt:
At least it's faster than the generic version (build with `--features lsx` vs without):

image

folkertdev:
Great numbers! 

> Awwww we can't run this in Miri quite yet

It's probably mostly a `stdarch` problem. E.g. https://github.com/rust-lang/stdarch/pull/214 fixes `lsx_vld`, for `lsx_vldi` that's harder but we can special-case the zero case in theory and just use `Maybeuninit::zeroed().assume_init()` there.

But then I just start running into missed optimizations, e.g. https://godbolt.org/z/qc8szMvWs, https://godbolt.org/z/Gv439esGf. That should be easy enough to fix but it's a bunch of LLVM work.

folkertdev:
Here is a version that Miri can run (but defeats much of the point currently because these implementations don't optimize well

https://gist.github.com/folkertdev/91f15d90143c32f9cd3bf4ed60189c59

Nevertheless, it shows what's needed in stdarch/LLVM, seems quite reasonable.

Gelbpunkt:
Yeah, I imagine with the ongoing work in stdarch, we will be able to run this in Miri at some point in the near future. For now, I'll remove `--features lsx` from the Miri run again.

Before I mark this as ready I also want to take one look at the whole `TAPS` thing, I imagine it should be trivial to precompute the extended `TAPS` indices and save on the `lsx_vexth_wu_hu`, etc. calls

folkertdev:
> Yeah, I imagine with the ongoing work in stdarch, we will be able to run this in Miri at some point in the near future. 

Well in many cases I think it does need LLVM work, and LLVM 23 is close so if that's missed we'd get it early 2027 at the earliest. So, not that near but it should be possible eventually.

Gelbpunkt:
> Before I mark this as ready I also want to take one look at the whole `TAPS` thing, I imagine it should be trivial to precompute the extended `TAPS` indices and save on the `lsx_vexth_wu_hu`, etc. calls

Implemented, here's the Godbolt diff: https://rust.godbolt.org/z/McY59fxbh

Performance-wise it doesn't seem to have an impact, we're trading a `lsx_vsllwil_wu_hu` + `lsx_vexth_wu_hu` for one load that is probably from cache. Went with this since it formats the code nicer.

heiher:
> lsx_vldi

portable impl: https://github.com/heiher/stdarch/commit/745b4fc578f19f338fe9c5bbfcd8da49b1312b09

PR #534, opened 2026-06-29
add `deflateUsed`

folkertdev:
Add `deflateUsed`. This is not implemented in `zlib-ng`, but it is in (new, december 2025) stock zlib.

### DIFF
--- a/zlib-rs/src/deflate.rs
+++ b/zlib-rs/src/deflate.rs
@@ -650,8 +650,9 @@ pub fn copy<'a>(
     };
 
     let mut bit_writer = BitWriter::from_pending(pending);
-    bit_writer.bits_valid = source_state.bit_writer.bits_valid;
     bit_writer.bit_buffer = source_state.bit_writer.bit_buffer;
+    bit_writer.bits_valid = source_state.bit_writer.bits_valid;
+    bit_writer.bits_used = source_state.bit_writer.bits_used;
 
     let dest_state = State {
         status: source_state.status,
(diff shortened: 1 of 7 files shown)


zlib-rs/src/deflate.rs
```diff
@@ -972,25 +986,29 @@ impl<'a> BitWriter<'a> {
     }
 
     fn flush_bits(&mut self) {
-        debug_assert!(self.bits_used <= 64);
-        let removed = self.bits_used.saturating_sub(7).next_multiple_of(8);
-        let keep_bytes = self.bits_used / 8; // can never divide by zero
+        debug_assert!(self.bits_valid <= 64);
+        let removed = self.bits_valid.saturating_sub(7).next_multiple_of(8);
+        let keep_bytes = self.bits_valid / 8; // can never divide by zero
```
bjorn3, line 991: Was the name previously incorrect?

folkertdev: No, I think zlib-ng renamed it from valid to used? Now we're using the stock zlib names.

bjorn3: Valid and used are the exact opposite values. If I understand correctly there would be an invariant `bits_valid + bits_used == 8` for everything except possibly the last byte.

folkertdev: `bits_valid` can be higher than 8 (all the way up to 56 I believe). It is the number of bits that are in the bit buffer, waiting for writing to the actual output.

`bits_used` is how many bits were actually written in the last byte that was written. It defaults to 8, but will calculate it in `emit_align`.

bjorn3: Why did zlib-ng rename it if it means something completely different?

folkertdev: I don't know. zlib-ng does not yet support these new additions to the API. 

In terms of the implementation we're not really beholden to what they do. If you think another name would be better, I'm open to suggestions.

bjorn3 (approved):

Issue #536, opened 2026-07-01
MSRV not upheld in 0.6.4 release


djc:
rustls has a 1.85 MSRV right now on `main`:

```
error[E0658]: `#[target_feature(..)]` can only be applied to `unsafe` functions
   --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/zlib-rs-0.6.4/src/adler32/neon.rs:119:5
    |
119 |     #[target_feature(enable = "neon")]
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
120 |     #[inline]
121 |     fn swap_64bit_lanes_when_be(v: uint8x16_t) -> uint8x16_t {
    |     -------------------------------------------------------- not an `unsafe` function
    |
    = note: see issue #69098 <https://github.com/rust-lang/rust/issues/69098> for more information
```

zlib-rs seems to advertise 1.75 but it doesn't seem to work at least in this particular case.

djc:
CI checks MSRV, but only on x86_64-unknown-linux-gnu. Probably good to at least add some kind of ARM target?

folkertdev:
Right, we check for more targets now, and I've release 0.6.5 which fixes the MSRV issue.

That said, it kind of demonstrates we really benefit from target features 1.1, stable since 1.86. I believe rustls also wanted to do an MSRV bump somewhat soon, what version are you aiming for?

djc:
I'd like to stick with 1.85 for the first few 0.24.x releases. But since zlib-rs use is optional in rustls, I think it'd probably be okay for zlib-rs to bump to a newer version, especially when they're semver-compatible bumps (such that we can stick with an MSRV-compatible version in CI but downstreams with a newer MSRV can implicitly pick up your newer versions).

djc:
This is from our [MSRV policy](https://github.com/rustls/rustls/blob/main/SECURITY.md#minimum-supported-rust-version):

<img width="819" height="156" alt="Image" src="https://github.com/user-attachments/assets/5b83c272-ba2e-4440-b2ab-938b9e966302" />

folkertdev:
Right, good to know it won't be a problem for you. And thanks for reporting this issue!

PR #538, opened 2026-07-01
check MSRV on more targets

folkertdev:
cc https://github.com/trifectatechfoundation/zlib-rs/issues/536

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -235,6 +235,40 @@ jobs:
         run: |
           cargo doc --target ${{matrix.target}} -p libz-rs-sys
 
+  check-msrv:
+    name: MSRV check (${{ matrix.target }})
+    runs-on: ubuntu-latest
+
+    strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/checks.yaml
```diff
@@ -235,6 +235,40 @@ jobs:
         run: |
           cargo doc --target ${{matrix.target}} -p libz-rs-sys
 
+  check-msrv:
```
bjorn3, line 238: Is the

```yaml
          - rust: msrv
            target: "x86_64-unknown-linux-gnu"
            os: ubuntu-latest
            codecov: false
```

above still necessary with this separate MSRV job?

folkertdev: That one actually runs the tests, which independently also seems useful?

bjorn3: Right. Didn't notice this was only a check build.

bjorn3 (approved):

PR #539, opened 2026-07-01
fix unsound return lifetime on Weak*Mut::as_mut_slice()

Manishearth:
This returns a mutable borrow while relinquishing mutable access to the source type. This is prone to lifetime extension bugs, but it's internal so it doesn't actually matter.

Worth fixing anyway.

(This issue was discovered by an audit performed using Gemini)


folkertdev:
Nice, thanks!

PR #549, opened 2026-07-28
Remove redundant casts

FranciscoTGouveia:
I believe these casts can be safely removed, as the source type is already the target type.

For reference, these changes were found by an optimizer developed as a research project.

CC @nunoplopes


folkertdev:
Neat. As I think you've found out one problem with these aliases that we run into is that they resolve to different types on different targets, so even if a conversion is redundant on one, it might be required on other targets.

We spoke at RustWeek right? We're starting to work on https://github.com/trifectatechfoundation/libzstd-rs-sys right now, which I think has a lot more low-hanging fruit for automatic cleanup.

folkertdev:
You can open an issue there or email me if you want to coordinate.

PR #552, opened 2026-08-02
Fix use-after-free in `set_level`

### LINKED ISSUE #551: Use-after-free introduced by `Deflate::compress_uninit`
### Summary
`Deflate::compress_uninit` stores input and output buffer as raw pointer into `Deflate`.
The buffers' lifetimes are untied, so input and output buffers can be dropped while `Deflate` is still alive.
Further access to `Deflate` may trigger use-after-free.

### Details
`Deflate::compress_uninit` converts input and output buffer into raw pointers and stores them in `Deflate.inner.next_in` and `Deflate.inner.next_out` respectively.
(`Deflate.inner` here has `DeflateStream` type)
https://github.com/trifectatechfoundation/zlib-rs/blob/5a82af8a944d3b2d23e861c5438bac35bc40d16d/zlib-rs/src/stable.rs#L340-L365

Since `next_in` and `next_out` are raw pointers with untied lifetime, it is possible to drop the input and output buffers leaving `Deflate` alive.
https://github.com/trifectatechfoundation/zlib-rs/blob/5a82af8a944d3b2d23e861c5438bac35bc40d16d/zlib-rs/src/deflate.rs#L37-L51

After dropping input and output buffers, stored raw pointers `Deflate.inner.next_in` and `Deflate.inner.next_out` remain as dangling pointers.
Further access to these pointers can trigger use-after-free.

In current version (0.6.6), `Deflate::set_level` (introduced at 0.5.4) is the only safe API that accesses one of these dangling pointers (`Deflate.inner.next_out`) and actually triggers UAF.
https://github.com/trifectatechfoundation/zlib-rs/blob/5a82af8a944d3b2d23e861c5438bac35bc40d16d/zlib-rs/src/stable.rs#L388-L395

`stable.rs:389 -> deflate.rs:465 -> deflate.rs:2510 -> deflate.rs:465 -> deflate.rs:2819`
It eventually reaches `std::ptr::copy_nonoverlapping` here:
https://github.com/trifectatechfoundation/zlib-rs/blob/5a82af8a944d3b2d23e861c5438bac35bc40d16d/zlib-rs/src/deflate.rs#L2805-L2826

### PoC
```rust
#![forbid(unsafe_code)]
use std::mem::MaybeUninit;
use zlib_rs::{Deflate, DeflateFlush};

fn main() {
    let mut d = Deflate::new(1, true, 15);

    let input = vec![b'a'; 1 << 16];

    {
        let mut output = vec![MaybeUninit::<u8>::uninit(); 1 << 20];
        d.compress_uninit(&input, &mut output, DeflateFlush::NoFlush).unwrap();

        // same thing, using `Deflate::compress` instead
        // let mut output = vec![0u8; 1 << 20];
        // d.compress(&input, &mut output, DeflateFlush::NoFlush).unwrap();
    }
    // output is dropped
    // but d.inner.next_out still holds old pointer to output

    d.set_level(9).unwrap(); // triggers UAF
}
```

Run following command to get ASan UAF report:
`RUSTFLAGS="-Zsanitizer=address" cargo +nightly run --target x86_64-unknown-linux-gnu`

Tested on zlib-rs 0.5.4 (first affected version?), 0.6.6 (latest release) and latest commit on main branch (5a82af8a944d3b2d23e861c5438bac35bc40d16d).

I reported this issue to the security page first and they said it's ok to post this on public issue.

folkertdev:
fixes https://github.com/trifectatechfoundation/zlib-rs/issues/551

### DIFF
--- a/zlib-rs/src/stable.rs
+++ b/zlib-rs/src/stable.rs
@@ -361,6 +361,13 @@ impl Deflate {
         self.total_in += (self.inner.next_in as usize - start_in as usize) as u64;
         self.total_out += (self.inner.next_out as usize - start_out as usize) as u64;
 
+        // Clear these pointers so there can be no use after free.
+        self.inner.next_in = core::ptr::null_mut();
+        self.inner.next_out = core::ptr::null_mut();
+
+        self.inner.avail_in = 0;
+        self.inner.avail_out = 0;
+
         ret
     }
 
@@ -386,6 +393,13 @@ impl Deflate {
     /// compression level. Flushing the stream before calling this method
     /// ensures that the function will succeed on the first call.
     pub fn set_level(&mut self, level: i32) -> Result<Status, DeflateError> {
+        // Clear these pointers so there can be no use after free.
+        self.inner.next_in = core::ptr::null_mut();
+        self.inner.next_out = core::ptr::null_mut();
+
+        self.inner.avail_in = 0;
+        self.inner.avail_out = 0;
+
         match crate::deflate::params(&mut self.inner, level, Default::default()) {
             ReturnCode::Ok => Ok(Status::Ok),
             ReturnCode::StreamError => Err(DeflateError::StreamError),
--- a/test-libz-rs-sys/src/deflate.rs
+++ b/test-libz-rs-sys/src/deflate.rs
@@ -1,9 +1,8 @@
+use core::ffi::{c_char, c_int, c_uint, c_ulong, CStr};
 use std::{ffi::CString, mem::MaybeUninit};
 
 // we use the libz_sys but configure zlib-ng in zlib compat mode
 use libz_sys as libz_ng_sys;
-
-use core::ffi::{c_char, c_int, c_uint, c_ulong, CStr};
 
 use libz_rs_sys::{
     deflate, deflateEnd, deflateInit2_, inflate, inflateEnd, inflateInit2_, Z_DEFLATED, Z_FILTERED,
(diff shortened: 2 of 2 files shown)


zlib-rs/src/stable.rs
```diff
@@ -386,6 +393,13 @@ impl Deflate {
     /// compression level. Flushing the stream before calling this method
     /// ensures that the function will succeed on the first call.
     pub fn set_level(&mut self, level: i32) -> Result<Status, DeflateError> {
+        // Clear these pointers so there can be no use after free.
+        self.inner.next_in = core::ptr::null_mut();
+        self.inner.next_out = core::ptr::null_mut();
+
+        self.inner.avail_in = 0;
+        self.inner.avail_out = 0;
```
bjorn3, line 401: Does this ensure that the `params` call below returns an error if anything needs to be flushed?

folkertdev: it does, `params` calls `deflate`, which checks

```rust
    if stream.avail_out == 0 {
        let err = ReturnCode::BufError;
        stream.msg = err.error_message();
        return err;
    }
```

folkertdev: ah, and then because no input was consumed, but there is still input available, `params` itself will return `BufError`.

bjorn3 (approved):

zlib-rs/src/stable.rs
folkertdev: ah, no, actually, we hit this now returning `StreamError`

```rust
pub fn deflate(stream: &mut DeflateStream, flush: DeflateFlush) -> ReturnCode {
    if stream.next_out.is_null()
        || (stream.avail_in != 0 && stream.next_in.is_null())
        || (stream.state.status == Status::Finish && flush != DeflateFlush::Finish)
    {
        let err = ReturnCode::StreamError;
        stream.msg = err.error_message();
        return err;
    }
```

Anyway, we error as intended.

PR #563, opened 2026-09-12
feat(cpu_features): enable runtime CPU feature detection in #![no_std]

gheffern:
## Summary

This PR enables runtime SIMD feature detection on x86/x86_64 in `#![no_std]` environments, unlocking hardware-accelerated AVX2, BMI1/2, AVX-512, and PCLMULQDQ code paths for bare-metal, embedded, and custom-allocator targets without introducing external crate dependencies.

Previously, runtime detection relied on `std::is_x86_feature_detected!` behind `#[cfg(feature = "std")]`. In builds where default features are disabled (such as bare-metal targets or `default-features = false, features = ["rust-allocator"]`), all detection functions unconditionally returned `false`. This caused silent fallback to scalar implementations, sacrificing substantial throughput (~708 MB/s vs ~976 MB/s on modern x86 hardware).

## Motivation & Architecture

1. **Strict Zero Dependencies**:
   `zlib-rs` serves as an embeddable, audit-friendly drop-in replacement for `zlib` and `zlib-ng`. Instead of pulling in third-party crates (e.g. `cpufeatures`), this PR implements unified, zero-dependency CPUID detection via `core::arch`.

2. **Single-Path Verification (Preventing Code Rot)**:
   By using the consolidated `core::arch` detection path for both `std` and `no_std` builds on x86, normal CI (`cargo test`) directly exercises and tests the exact same detection logic used in `#![no_std]` environments.

3. **1:1 Parity & Scoped Diff**:
   The diff is strictly contained to a single file (`zlib-rs/src/cpu_features.rs`). All existing helper functions (including `is_enabled_sse42` and the `CpuFeatures` struct) and signatures from `main` are preserved to guarantee 100% backward and internal API compatibility.

## Technical Details

- **Single-Pass CPUID Detection**: Queries Leaf 0, Leaf 1, and Leaf 7 subleaf 0 once, caching discovered flags in a static `AtomicU32` with `Ordering::Relaxed`. Subsequent queries resolve via a single atomic bitwise check.
- **Microarchitectural Guardrails**:
  - Validates `cpuid0.eax >= 1` before querying Leaf 1, and `cpuid0.eax >= 7` before querying Leaf 7 subleaf 0 to prevent reading aliased data on processors or hypervisors with truncated leaf limits.
  - Verifies `OSXSAVE` (Leaf 1 ECX bit 27) before invoking `_xgetbv(0)`, preventing `#UD` (invalid opcode) faults on hypervisors or environments where OS state saving is disabled.
  - Validates OS state-saving masks in XCR0: `0x6` (XMM + YMM) for AVX2/BMI, and `0xe6` (XMM + YMM + opmask + ZMM_Hi256 + Hi16_ZMM) for AVX-512.
  - Gating for AVX-512 requires `AVX512F` (bit 16), `AVX512BW` (bit 30), and `AVX512VL` (bit 31) to prevent illegal instruction exceptions on partial-AVX512 microarchitectures.
  - Merged Leaf 1 ECX PCLMULQDQ (bit 1) and SSE4.1 (bit 19) checks.
- **Miri Compatibility**: Short-circuits with `return 0` under `#[cfg(miri)]` so virtual test runs avoid unsupported inline assembly aborts.

## Empirical Benchmark Impact (`default-features = false, features = ["rust-allocator"]`)

Tested on AMD Zen 4 across the Silesia corpus (20 iterations, physical core pinning):
- **Decompression Throughput**: **976.09 MB/s** (vs 708.95 MB/s unvectorized baseline = **+37.68% speedup**)
- **Compression Throughput**: **76.70 MB/s** (vs 74.73 MB/s baseline = **+2.64% speedup**)
- Bit-exact data integrity verified across all 12 Silesia files.

## CI & Compatibility Matrix

- `cargo fmt --all --check` (pass)
- `cargo clippy --workspace --all-targets -- -D warnings` (pass)
- `cargo test -p zlib-rs` (79 unit tests + 4 doc tests passed)
- `cargo test -p zlib-rs --no-default-features --features rust-allocator` (75 unit tests passed)
- `cargo test -p zlib-rs --no-default-features --features c-allocator` (73 unit tests passed)


zlib-rs/src/cpu_features.rs
```diff
@@ -8,113 +7,321 @@ impl CpuFeatures {
...
 #[inline(always)]
 pub fn is_enabled_sse() -> bool {
-    #[cfg(any(target_arch = "x86_64", target_arch = "x86"))]
-    #[cfg(feature = "std")]
-    return std::is_x86_feature_detected!("sse");
-
-    false
+    #[cfg(target_arch = "x86_64")]
+    {
+        true
```
bjorn3, line 108: SSE can be disabled on x86_64 too. Especially in no_std contexts. (I don't know if kernels actually disable it in kernel space, but it is definitively not allowed there on most kernels)

bjorn3:
Was this LLM generated? The PR description is rather verbose and contains some info that is useless like all things that CI would check anyway.

gheffern:
Yes it was. Largely via Gemini 3.8 Flash high.  I did review it all myself to the best of my ability. And tried to pare down the verboseness where it made sense to me, but it certainly keeps its AI generated style.

zlib-rs/src/cpu_features.rs
gheffern: If we updated is_enabled_sse like so:
```
   #[inline(always)]
    pub fn is_enabled_sse() -> bool {
        #[cfg(target_arch = "x86_64")]
        {
            cfg!(target_feature = "sse")
        }
        #[cfg(target_arch = "x86")]
        {
            cfg!(target_feature = "sse") || (x86_features::get() & x86_features::SSE) != 0
        }
        #[cfg(not(any(target_arch = "x86_64", target_arch = "x86")))]
        {
            false
        }
    }
```
And in x86_features::detect(), short-circuit to return 0 when cfg!(all(target_arch = "x86_64", not(target_feature = "sse"))) so that AVX2 / AVX-512 /
PCLMULQDQ runtime paths are also suppressed when SSE is stripped? 

Does that address this issue? Or not really, and this is just kind of a dead end idea?

